PR #3, opened 2019-05-13
Fix link to documentation

bryik:
The current link results in a 404.

The new link is based on [the link to Geocube's documentation](https://github.com/corteva/geocube/blob/master/README.rst) (which works).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,7 +4,7 @@ rioxarray README
 
 rasterio xarray extension.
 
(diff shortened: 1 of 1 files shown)


snowman2:
@bryik, many thanks for pointing this out! Instead of updating the docs as you proposed, I added a file to redirect users to the docs (similar to geocube) in this commit: https://github.com/corteva/rioxarray/commit/ba18c367b3e9617d925c5228f231020b426406b2

PR #8, opened 2019-06-25
added ability to export data array to raster

### LINKED ISSUE #6: Add ability to export to geotiff
It would be useful to write the dataframe out to a geotiff file.

snowman2:
- [x] Closes #6
 - [x] Tests added

### DIFF
--- a/rioxarray/rioxarray.py
+++ b/rioxarray/rioxarray.py
@@ -769,6 +769,41 @@ class RasterArray(XRasterBase):
 
         return interp_array
 
+    def to_raster(self, raster_path, driver="GTiff", dtype=None):
+        """
+        Export the DataArray to a raster file.
+
+        Parameters
+        ----------
+        raster_path: str
+            The path to output the raster to.
+        driver: str
+            The name of the GDAL/rasterio driver to use to export the raster.
+            Default is "GTiff".
+        dtype: str
+            The data type to write the raster to. Default is the datasets dtype.
+
+        """
+        width, height = self.shape
+        dtype = str(self._obj.dtype) if dtype is None else dtype
+        with rasterio.open(
+            raster_path,
+            "w",
+            driver=driver,
+            height=int(height),
+            width=int(width),
+            count=len(self._obj.dims),
+            dtype=dtype,
+            crs=self.crs,
+            transform=self.transform(recalc=True),
+            nodata=self.nodata,
+        ) as dst:
+            data = self._obj.values.astype(dtype)
+            if data.ndim == 2:
+                dst.write(data, 1)
+            else:
+                dst.write(data)
+
 
 @xarray.register_dataset_accessor("rio")
 class RasterDataset(XRasterBase):
--- a/rioxarray/rioxarray.py
+++ b/rioxarray/rioxarray.py
@@ -347,16 +347,16 @@ class RasterArray(XRasterBase):
         elif extra_dims and self._obj.dims != (extra_dims[0], self.y_dim, self.x_dim):
             raise InvalidDimensionOrder(
                 "Invalid dimension order. Expected order: {0}. "
-                "You can use `DataArray.transpose{0}` to reorder your dimensions.".format(
+                "You can use `DataArray.transpose{0}`"
+                " to reorder your dimensions.".format(
                     (extra_dims[0], self.y_dim, self.x_dim)
                 )
             )
         elif not extra_dims and self._obj.dims != (self.y_dim, self.x_dim):
             raise InvalidDimensionOrder(
                 "Invalid dimension order. Expected order: {0}"
-                "You can use `DataArray.transpose{0}` to reorder your dimensions.".format(
-                    (self.y_dim, self.x_dim)
-                )
+                "You can use `DataArray.transpose{0}` "
+                "to reorder your dimensions.".format((self.y_dim, self.x_dim))
             )
         return extra_dims[0] if extra_dims else None
 
--- a/rioxarray/rioxarray.py
+++ b/rioxarray/rioxarray.py
@@ -346,13 +346,17 @@ class RasterArray(XRasterBase):
             raise TooManyDimensions("Only 2D and 3D data arrays supported.")
         elif extra_dims and self._obj.dims != (extra_dims[0], self.y_dim, self.x_dim):
             raise InvalidDimensionOrder(
-                "Invalid dimension order. Expected order: {}".format(
+                "Invalid dimension order. Expected order: {0}. "
+                "You can use `DataArray.transpose{0}` to reorder your dimensions."
+                .format(
                     (extra_dims[0], self.y_dim, self.x_dim)
                 )
             )
         elif not extra_dims and self._obj.dims != (self.y_dim, self.x_dim):
             raise InvalidDimensionOrder(
-                "Invalid dimension order. Expected order: {}".format(
+                "Invalid dimension order. Expected order: {0}"
+                "You can use `DataArray.transpose{0}` to reorder your dimensions."
+                .format(
                     (self.y_dim, self.x_dim)
                 )
             )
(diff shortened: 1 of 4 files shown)


snowman2:
Appveyor builds not working:
```
Traceback (most recent call last):
  File "C:\Miniconda3-x64\lib\site-packages\conda\exceptions.py", line 1043, in __call__
    return func(*args, **kwargs)
  File "C:\Miniconda3-x64\lib\site-packages\conda\cli\main.py", line 84, in _main
    exit_code = do_call(args, p)
  File "C:\Miniconda3-x64\lib\site-packages\conda\cli\conda_argparse.py", line 80, in do_call
    module = import_module(relative_mod, __name__.rsplit('.', 1)[0])
  File "C:\Miniconda3-x64\lib\importlib\__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
  File "<frozen importlib._bootstrap>", line 983, in _find_and_load
  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "C:\Miniconda3-x64\lib\site-packages\conda\cli\main_create.py", line 10, in <module>
    from .install import install
  File "C:\Miniconda3-x64\lib\site-packages\conda\cli\install.py", line 19, in <module>
    from ..core.index import calculate_channel_urls, get_index
  File "C:\Miniconda3-x64\lib\site-packages\conda\core\index.py", line 9, in <module>
    from .package_cache_data import PackageCacheData
  File "C:\Miniconda3-x64\lib\site-packages\conda\core\package_cache_data.py", line 15, in <module>
    from conda_package_handling.api import InvalidArchiveError
  File "C:\Miniconda3-x64\lib\site-packages\conda_package_handling\api.py", line 3, in <module>
    from libarchive.exception import ArchiveError as _LibarchiveArchiveError
  File "C:\Miniconda3-x64\lib\site-packages\libarchive\__init__.py", line 1, in <module>
    from .entry import ArchiveEntry
  File "C:\Miniconda3-x64\lib\site-packages\libarchive\entry.py", line 6, in <module>
    from . import ffi
  File "C:\Miniconda3-x64\lib\site-packages\libarchive\ffi.py", line 27, in <module>
    libarchive = ctypes.cdll.LoadLibrary(libarchive_path)
  File "C:\Miniconda3-x64\lib\ctypes\__init__.py", line 434, in LoadLibrary
    return self._dlltype(name)
  File "C:\Miniconda3-x64\lib\ctypes\__init__.py", line 356, in __init__
    self._handle = _dlopen(self._name, mode)
TypeError: LoadLibrary() argument 1 must be str, not None
During handling of the above exception, another exception occurred:
```

djhoese:
Would it be possible to also add a "tags" keyword argument?

snowman2:
> Would it be possible to also add a "tags" keyword argument?

That could be useful. Do you have an example of how you would like this to work?

djhoese:
In `trollimage` we use it for the image DATETIME by default: https://github.com/pytroll/trollimage/blob/master/trollimage/xrimage.py#L340

Otherwise, we have a lot of users who have custom visualization clients or other tools where they've historically had to throw extra metadata in to tiff tags/metadata (@mraspaud may have examples). But a pretty common one is people adding a `scale_factor` and `add_offset` to their metadata so the visualization client can use them to get "real" data values back.

snowman2:
> Oh and what about a colorinterp?

Maybe a default `**kwargs` you can pass into writing the raster would be the solution here. I have been thinking about this and there are many options that would be nice to add.

djhoese:
> Maybe a default **kwargs you can pass into writing the raster would be the solution here. 

Yeah I didn't want to suggest it since most people get annoyed of "catch all" kwargs if they can be avoided.

snowman2:
> Yeah I didn't want to suggest it since most people get annoyed of "catch all" kwargs if they can be avoided.

I definitely understand that. I have been debating it in my head a bit. The catch all is annoying, but if something is added in another version of rasterio or a driver specific option is added and it is not explicitly set here, the function won't be able to handle it. Another option is to add an argument for a rasterio profile dictionary.

snowman2:
Looks like rasterio.open uses kwargs, so I don't think it would be terrible to use kwargs here as well:
https://github.com/mapbox/rasterio/blob/fc9861cef7a70f5136dd4b47920d05ea88312137/rasterio/__init__.py#L58

snowman2:
I think adding a `tags` dict option that would be called in `update_tags` would solve your problems, correct?
https://github.com/pytroll/trollimage/blob/63fa32f2d40bb65ebc39c4be1fb1baf8f163db98/trollimage/xrimage.py#L357

djhoese:
It would make it more flexible for users, yes. I forgot about the `update_tags` call being needed.

snowman2:
@djhoese I am planning to merge these changes in soon as the tests pass (for non-windows builds at least). See any problems with the code in its current state?

rioxarray/rioxarray.py
```diff
@@ -327,6 +332,32 @@ def _int_bounds(self):
...
+        they are are in the proper order.
+
+        Returns:
+        --------
+        str or None: Name extra dimension.
+        """
+        extra_dims = list(set(list(self._obj.dims)) - set([self.x_dim, self.y_dim]))
+        if len(extra_dims) > 1:
+            raise TooManyDimensions("Only 2D and 3D data arrays supported.")
+        elif extra_dims and self._obj.dims != (extra_dims[0], self.y_dim, self.x_dim):
```
djhoese: This check seems odd to me since it depends on the order of `extra_dims`. If the data is 4D (which is not checked here) then the "bands" dimension could be anywhere. Also since you converted the dims to sets and then back to lists you don't know the order of the extra dims.

Do you have anything that defines the name of the "bands" dimension? Sorry, not super familiar with rioxarray.

djhoese: You could use xarray's `transpose` method to shift dimensions around too.

snowman2: > This check seems odd to me since it depends on the order of extra_dims. 

Currently it is okay since the dataset is ensured to have only 2 or 3 dimensions. So, there should only be up to 1 extra dimension.

> If the data is 4D (which is not checked here) then the "bands" dimension could be anywhere. 

Currently 4D data is not supported or allowed. But, it would be nice to eventually reach n-dim support.

> Do you have anything that defines the name of the "bands" dimension?

If the dataset is 2D, then there won't be a bands dimension and it is just 1 band. But, if the dataset is 3D, then it will always be the dimension that is not the `x` or `y` dimension.

snowman2: > You could use xarray's transpose method to shift dimensions around too.

That is a really nice option. I have been thinking about how to handle that as I am not sure how I feel about it being magically handled. For example, when reading the dataset back in from a tif using `open_rasterio`, the dimension order would be swapped and the user wouldn't know it. Maybe recommending swapping the dimension order using `DataArray.transpose` in the error message would be useful. 

Thoughts?

djhoese: Yeah, this library is lower level than `trollimage` which is where most of my rasterio writing experience comes from. I think in this case less magic behavior is probably better.

As for 2D and 3D being the only ones allowed, is that check done outside of this method? The top if statement in this PR only checks for more than 1 dimension I thought.

djhoese: 🤦‍♂ Ah you are checking that there is only 1 extra dimension. Got it.

snowman2:
```python
extra_dims = list(set(list(self._obj.dims)) - set([self.x_dim, self.y_dim]))
```
In this case, the extra_dims are the dimensions not in the expected `x` and `y` dimensions. So, if it does not have an extra dimension, it is 2D. If it has 1 extra dimension it is 3D, etc ...

snowman2:
> The only other helpful long term solution would be to use windowed writing like we do in trollimage to achieve better performance with dask (saving chunk by chunk), but this has its own issues we've discovered with GDAL's in-memory caching and making large output files if you don't have a lot of memory available.

That is a good idea. With the issues you are mentioning, I think I will create a separate issue/PR for that to better understand the issues.

snowman2:
Thanks for the review @djhoese!

PR #14, opened 2019-06-27
Added methods to set and write CRS

### LINKED ISSUE #5: Add ability to set CRS and write CRS information to dataset
See: https://github.com/geoxarray/geoxarray/issues/9

snowman2:
- [x] Closes #5 
 - [x] Tests added

### DIFF
--- a/rioxarray/rioxarray.py
+++ b/rioxarray/rioxarray.py
@@ -25,6 +25,7 @@ from scipy.interpolate import griddata
 
 from rioxarray.exceptions import (
     InvalidDimensionOrder,
+    MissingCRS,
     NoDataInBounds,
     OneDimensionalRaster,
     TooManyDimensions,
@@ -32,7 +33,7 @@ from rioxarray.exceptions import (
 from rioxarray.crs import crs_to_wkt
 
 FILL_VALUE_NAMES = ("_FillValue", "missing_value", "fill_value", "nodata")
-UNWANTED_RIO_ATTRS = ("nodata", "nodatavals", "crs", "is_tiled", "res")
+UNWANTED_RIO_ATTRS = ("nodatavals", "crs", "is_tiled", "res")
 DEFAULT_GRID_MAP = "spatial_ref"
 
 
@@ -123,23 +124,9 @@ def add_xy_grid_meta(coords):
 
 
 def add_spatial_ref(in_ds, dst_crs, grid_map_name):
-    # remove old grid map if exists
-    try:
-        del in_ds.coords[grid_map_name]
-    except KeyError:
-        pass
-
-    # add grid mapping variable
-    in_ds.coords[grid_map_name] = xarray.Variable((), 0)
-    match_proj = crs_to_wkt(CRS.from_user_input(dst_crs))
-
-    grid_map_attrs = dict()
-    # add grid mapping variable
-    grid_map_attrs["spatial_ref"] = match_proj
-    # http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#appendix-grid-mappings
-    # http://desktop.arcgis.com/en/arcmap/10.3/manage-data/netcdf/spatial-reference-for-netcdf-data.htm
-    grid_map_attrs["crs_wkt"] = match_proj
-    in_ds.coords[grid_map_name].attrs = grid_map_attrs
+    in_ds.rio.write_crs(
+        input_crs=dst_crs, grid_mapping_name=grid_map_name, inplace=True
+    )
     return in_ds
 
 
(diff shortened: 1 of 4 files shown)


rioxarray/rioxarray.py
```diff
@@ -230,7 +216,79 @@ def crs(self):
         """:obj:`rasterio.crs.CRS`:
             The projection of the dataset.
         """
-        return self._crs
+        raise NotImplementedError
+
+    def set_crs(self, input_crs):
```
djhoese: So you decided not to use the `@property` version of this?

snowman2: I think having it as a method could potentially be more flexible, allow method chaining, and it would match the interface in `geopandas`. See: https://github.com/geopandas/geopandas/pull/747#issuecomment-398913889

Thoughts?

djhoese: Ok, but it seems like most of their argument is related to method chaining and your `set_crs` doesn't return anything. That said, I like the idea of consistency.

rioxarray/rioxarray.py
```diff
@@ -230,7 +216,79 @@ def crs(self):
...
+        input_crs: object
+            Anythong accepted by `rasterio.crs.CRS.from_user_input`.
+        """
+        self._crs = CRS.from_user_input(input_crs)
+
+    def write_crs(
+        self, input_crs=None, grid_mapping_name=DEFAULT_GRID_MAP, inplace=False
+    ):
+        """
+        Write the CRS to the dataset in a CF compliant manner.
```
djhoese: Is there a reason CF is the only available option for such a generic method name (`write_crs`)? Do any of the existing methods assign the CRS object to a `crs` coordinate?

rioxarray/rioxarray.py
snowman2: > set_crs doesn't return anything

Oh, hah :man_facepalming:. Good catch. I should fix that. Thanks!

rioxarray/rioxarray.py
snowman2: > Is there a reason CF is the only available option for such a generic method name

It is currently the only method supported by `rioxarray` for writing the CRS to the dataset. Did you have other conventions you were thinking of following for writing?

> Do any of the existing methods assign the CRS object to a crs coordinate?

You can assign it to the CRS coordinate using the `grid_mapping_name`. You can see an example of this in the tests in the PR.

rioxarray/rioxarray.py
snowman2: Chaining should work now.

djhoese: How does this work with xarray accessors? You're returning the xarray object, but does that object have this same `_crs` attribute that was just set? To match geopandas does this need to set the CRS on a copy of the xarray object and then return that copy?

rioxarray/rioxarray.py
djhoese: > Did you have other conventions you were thinking of following for writing?

No, you're right. CF is really the only "new" format that needs conversion I guess. Except maybe pyproj CRS -> rasterio CRS. I guess I was just mainly noticing that this method writes to the xarray object associated with it and I could see that being useful for things like setting a `crs` coordinate or adding `y` and `x` coordinates. Likely something more in geoxarray's "generic" realm though.

rioxarray/rioxarray.py
```diff
@@ -255,17 +320,23 @@ def crs(self):
             Retrieve projection from `xarray.DataArray`
         """
         if self._crs is not None:
-            return self._crs
+            return None if self._crs is False else self._crs
```
djhoese: When is `self._crs` False?

rioxarray/rioxarray.py
snowman2: Ah, you want to attach the `CRS` object to the coordinate instead of the CF parameters to the coordinate. No, that is currently not supported.

rioxarray/rioxarray.py
snowman2: It is `False` when it has gone through the process of looking for the CRS in all of the expected locations and failed. This prevents it from re-checking every time when the CRS is not found from attributes in the dataset.

djhoese: Nice.

rioxarray/rioxarray.py
djhoese: Yes, that was my hope for geoxarray (`data_arr.coords['crs'].item()` is a pyproj.crs.CRS object). Maybe I'll finally get the time to do it during SciPy...yeah right.

snowman2: > data_arr.coords['crs'].item() is a pyproj.crs.CRS object

Just curious, what is the benefit of that versus having the property `geo.crs`? I guess it does come into play with checking the equality of two DataArrays. Any other reasons? My only concern with that would be that it would make it so the dataset is no longer serializable and cannot be written with `to_netcdf`. But. I guess you could add `geo.to_netcdf`.
(maybe this is a discussion that should be moved to a separate issue in geoxarray?).

> Maybe I'll finally get the time to do it during SciPy

Well, you can always hope :crossed_fingers: 

> Except maybe pyproj CRS -> rasterio CRS

I think this is something that should be handled in `set_crs` (which will be called by `write_crs`).

djhoese: What determines the serialization of this? I suppose it would have to be handled in core xarray. Yes, this isn't a great situation. The goal was to have xarray treat the CRS as a "real" coordinate, but I later found out that xarray doesn't actually care if the CRS matches or not if it isn't a dimensional coordinate (like `y` or `x`) and will end up just dropping the `crs` coordinate if needed.

I wasn't a huge fan of `data_arr.geo.crs` because it felt "hidden" and seemed like it would be difficult to explain why this valuable information has to be accessed this way (thinking from a tutorial/teaching point of view).

snowman2: Here is the serialization issue I was talking about:
```python
>>> import xarray
>>> from rasterio.crs import CRS
>>> xds = xarray.Dataset()
>>> xds.coords["crs"] = xarray.Variable((), CRS.from_epsg(4326))
>>> xds
<xarray.Dataset>
Dimensions:  ()
Coordinates:
    crs      object <bound method Mapping.values of CRS.from_epsg(4326)>
Data variables:
    *empty*
>>> xds.to_netcdf("test_crs.nc")
...
ValueError: unable to infer dtype on variable 'crs'; xarray cannot serialize arbitrary Python objects
```

> I wasn't a huge fan of data_arr.geo.crs because it felt "hidden"  ...

I guess it is not really visible on the xarray dataset. I think if there is documentation and examples of proper usage, it has potential to be an okay path forward.

snowman2: I am going to mark this as resolved as the conversation has moved to something that should be in another issue.

snowman2:
@djhoese any other things you have noticed that should be addressed before merging?

rioxarray/rioxarray.py
snowman2: > You're returning the xarray object, but does that object have this same _crs attribute that was just set?

It seems to be the case in the tests.

```python
>>> test_da = xarray.DataArray(
...     numpy.zeros((5, 5)),
...     dims=("y", "x"),
...     coords={"y": numpy.arange(1, 6), "x": numpy.arange(2, 7)},
... )
>>> test_da.rio.crs
>>> ds2 = test_da.rio.set_crs(4326)
>>> ds2.rio.crs
CRS.from_epsg(4326)
>>> test_da.rio.crs
CRS.from_epsg(4326)

```

> To match geopandas does this need to set the CRS on a copy of the xarray object and then return that copy?

Hmmm, I guess an `inplace=` argument would be useful. It seems the default they are thinking of is `inplace=True` as making a copy when setting the attribute is not entirely necessary.

snowman2: Just made the change.

djhoese:
Last question, with the `set_crs` it returns a copy of the `self._obj` but does the original `self._obj.rio.crs` also get set? IMO only the *returned* xarray object should have `crs` set, right?

djhoese:
🤦‍♂ It would help if I looked at your most recent commits. All looks good. Nice job.

snowman2:
Thanks for the reviews @djhoese and @alfredoahds!

Issue #20, opened 2019-07-09
Nodata from `xarray.open_rasterio` is always NaN


snowman2:
Nodata from `xarray.open_rasterio` is always NaN. This is problematic with integer data (writing to_raster) or for clarity purposes.

PR #23, opened 2019-07-10
use input raster profile to write output raster profile & preserve None nodata if opened with xarray.open_rasterio

### LINKED ISSUE #19: In to_raster, check for original raster object profile for writing to raster
image

snowman2:
- [x] Closes #19 & #20
 - [x] Tests added

### DIFF
--- a/rioxarray/rioxarray.py
+++ b/rioxarray/rioxarray.py
@@ -755,7 +755,7 @@ class RasterArray(XRasterBase):
 
         return cl_array
 
-    def clip(self, geometries, crs, all_touched=False):
+    def clip(self, geometries, crs, all_touched=False, drop=True):
         """
         Crops a :class:`xarray.DataArray` by geojson like geometry dicts.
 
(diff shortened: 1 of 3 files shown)


rioxarray/rioxarray.py
```diff
@@ -375,11 +375,16 @@ def nodata(self):
...
-            except (KeyError, IndexError):
-                pass
+                self._nodata = self._obj._file_obj.acquire().nodata
+            except AttributeError:
+                try:
+                    self._nodata = self._obj.attrs["nodatavals"][0]
+                except (KeyError, IndexError):
+                    pass
 
         if self._nodata is None:
```
alfredoahds: I know this hasn't changed in this PR, but are there any concerns with setting nodata to False? What's the benefit over None?

snowman2: Setting it to False make it so that the entire process for looking for the nodata value wont have to be re-run through each time in the case that the value is None. See the beginning of the method.

snowman2: https://github.com/corteva/rioxarray/blob/8605b618c696a24271f8c87ea405a463364f602f/rioxarray/rioxarray.py#L364-L366

snowman2:
Just FYI, I had the other MR pointing to this one for visualizing purposes and forgot I did it and merged it in (need to remember not to do this in the future :)) - so this PR just got a little messier.

PR #26, opened 2019-07-11
 added 'drop' argument for clip; fixed geometry crs reprojection in clip 

### LINKED ISSUE #25: add `drop` option to `clip()`
Looks like the clip function in QGIS defaults to match the extent of the masked geometry:
image

But, it also has the option to keep the original resolution of the original raster. I am thinking about adding the `drop` argument to the `clip()` function to preserve the original raster an only return a masked version.

Related to ihttps://github.com/pydata/xarray/issues/501

snowman2:
- [x] Closes #25
 - [x] Tests added

### DIFF
--- a/rioxarray/rioxarray.py
+++ b/rioxarray/rioxarray.py
@@ -755,7 +755,7 @@ class RasterArray(XRasterBase):
 
         return cl_array
 
-    def clip(self, geometries, crs, all_touched=False):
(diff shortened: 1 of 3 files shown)


snowman2:
Thanks @alfredoahds!

snowman2:
Whoops, forgot this one was set to merge into the other branch.

PR #31, opened 2019-07-23
added rioxarray.open_rasterio

### LINKED ISSUE #7: Add open function.
The `xarray.open_rasterio` function exists, but it has issues with loss of CRS information due to export to PROJ string. Additionally, it would be nice to have the data loaded into standard netCDF locations (e.g. `nodatavals` should use `_FillValue`).

snowman2:
- [x] Closes #7
 - [x] Tests added

Changes from xarray version:
- Added CRS as WKT in standard location (https://github.com/pydata/xarray/issues/2723)
- Add nodata in standard location (https://github.com/pydata/xarray/issues/2308)
- Add mask option (https://github.com/pydata/xarray/pull/2618)
- Added automatic chunking (https://github.com/pydata/xarray/pull/2255)


rioxarray/_io.py
```diff
@@ -0,0 +1,392 @@
...
+
+from rioxarray.rioxarray import affine_to_coords
+
+
+# TODO: should this be GDAL_LOCK instead?
+RASTERIO_LOCK = SerializableLock()
+
+_ERROR_MSG = (
+    "The kind of indexing operation you are trying to do is not "
+    "valid on rasterio files. Try to load your data with ds.load()"
```
justingruca: Extra space needed at the end here.

rioxarray/_io.py
```diff
@@ -0,0 +1,392 @@
...
+        band_key: an indexer for the 1st dimension
+        window: two tuples. Each consists of (start, stop).
+        squeeze_axis: axes to be squeezed
+        np_ind: indexer for loaded numpy array
+
+        See also
+        --------
+        indexing.decompose_indexer
+        """
+        assert len(key) == 3, "rasterio datasets should always be 3D"
```
justingruca: This might be worth a real exception.

snowman2:
Need new doc page ...

rioxarray/_io.py
```diff
@@ -0,0 +1,392 @@
...
+        store multiple times. Defaults to True unless you specify the `chunks`
+        argument to use dask, in which case it defaults to False.
+    lock : False, True or threading.Lock, optional
+        If chunks is provided, this argument is passed on to
+        :py:func:`dask.array.from_array`. By default, a global lock is
+        used to avoid issues with concurrent access to the same file when using
+        dask's multithreaded backend.
+    masked : bool, optional
+        If True, read the mask and to set values to NaN. Defaults to False.
+return
```
justingruca: Remove this?

rioxarray/_io.py
```diff
@@ -0,0 +1,392 @@
...
+    coords["band"] = np.asarray(riods.indexes)
+
+    # Get coordinates
+    if LooseVersion(rasterio.__version__) < "1.0":
+        transform = riods.affine
+    else:
+        transform = riods.transform
+    if transform.is_rectilinear:
+        # 1d coordinates
+        parse = True if parse_coordinates is None else parse_coordinates
```
justingruca: Is `parse` supposed to be used somewhere other than the `if` statement below? If not maybe just fold the assignment logic into the `if` statement. Same below for 2d coordinates.

rioxarray/_io.py
```diff
@@ -0,0 +1,392 @@
...
+            mtime = os.path.getmtime(filename)
+        except OSError:
+            # the filename is probably an s3 bucket rather than a regular file
+            mtime = None
+
+        if chunks in (True, "auto"):
+            from dask.array.core import normalize_chunks
+            import dask
+
+            if dask.__version__ < "0.18.0":
```
justingruca: This doesn't work for Dask versions where the major number is 0 and the minor number is a single digit, although that may not matter.

snowman2:
Seems to work pretty well:
image

rioxarray/_io.py
snowman2: Good catch. Thanks!

rioxarray/_io.py
snowman2: Good catch, I agree.

rioxarray/_io.py
snowman2: Okay, I will use LooseVersion here.

rioxarray/_io.py
snowman2: Agreed!

rioxarray/_io.py
snowman2: Probably just remove. It appears unused.

PR #34, opened 2019-08-01
added function to add 2D latlon to dataset

snowman2:
- [x] Tests added


djhoese:
Does this still work if someone has a `+proj=latlong` projection and 1D longitude and latitude coordinates that they use as their X and Y coordinates?

snowman2:
> Does this still work if someone has a +proj=latlong projection and 1D longitude and latitude coordinates that they use as their X and Y coordinates?

It should. As long as they don't have 'latitude' and 'longitude' in the coordinates it will generate a 2D array of latitude and longitude.

PR #36, opened 2019-08-07
added support for custom CRS with wkt attribute for datacube CRS support

### LINKED ISSUE #35: Add support for initializing from datacube CRS
https://github.com/opendatacube/datacube-core/blob/develop/datacube/utils/geometry/_base.py

Currently it raises an exception as it cannot initialize from it. Thinking about checking for the `crs_str` attribute.

snowman2:
- [x] Closes #35
 - [x] Tests added


snowman2:
Note - commits will be squashed on merge.

PR #42, opened 2019-09-03
added rio.set_nodata,rio.write_nodata,rio.set_attrs,rio.update_attrs

### LINKED ISSUE #37: Add set_attrs/update_attrs methods (or an equivalent)
Currently, you could set the nodata or crs metadata in the attrs, but it won't be re-discovered if you already attempted to access them (reguardless if it was found or not). These methods would reset the cache properly so the accessor knows to re-search for these properties.

snowman2:
- [x] Closes #37
 - [x] Tests added

### DIFF
--- a/rioxarray/rioxarray.py
+++ b/rioxarray/rioxarray.py
@@ -15,14 +15,14 @@ from abc import abstractmethod
 from datetime import datetime
 
 import numpy as np
+
 import rasterio.warp
 import xarray
 from affine import Affine
 from rasterio.crs import CRS
 from rasterio.enums import Resampling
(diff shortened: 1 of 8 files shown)


rioxarray/rioxarray.py
```diff
@@ -238,6 +255,8 @@ def set_crs(self, input_crs, inplace=True):
         ----------
         input_crs: object
             Anythong accepted by `rasterio.crs.CRS.from_user_input`.
```
justingruca: Not part of the diff, but there's a typo here.

snowman2: Nice catch.

PR #50, opened 2019-09-23
Use xarray.open_rasterio for rioxarray.open_rasterio with xarray<0.12.3

snowman2:


snowman2:
Thanks for reviewing @justingruca!

PR #51, opened 2019-09-27
improve CRS searching for xarray.Dataset & use default grid mapping name

### LINKED ISSUE #44: Drop python 2 support
It will be removed in rasterio, fiona. Already dropped in pyproj.

snowman2:
- [x] Tests added

Also closes #44. Snuck it in because using several python 3 only things.


snowman2:
This is helpful for when the `grid_mapping` attribute is lost but the default `spatial_ref` is still in the `coords`.

snowman2:
Unrelated MacOS python 3.6 failure

PR #54, opened 2019-10-24
write to raster using windowed writing & add isel_window method

### LINKED ISSUE #22: Add ability to get raster subset with a rasterio Window
Could be nice to support that in the interface

snowman2:
- [x] Closes #22 (start of #9)
 - [x] Tests added

The thing that I am on the fence about is if we want to make windowed writing the default or the previous behavior the default. The windowed writing can be more memory efficient, but may not be as fast reading the entire chunk into memory. There are different scenarios. So, maybe add a `windowed=True` kwarg to turn on/off this behavior.

Side note: since the block sizes of the destination raster are created with rasterio, it should be constant across all bands, so that isn't something to worry about.

### DIFF
--- a/rioxarray/rioxarray.py
+++ b/rioxarray/rioxarray.py
@@ -60,8 +60,8 @@ def affine_to_coords(affine, width, height, x_dim="x", y_dim="y"):
     dict: x and y coordinate arrays.
 
     """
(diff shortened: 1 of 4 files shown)


snowman2:
Also, added fix for:
```
DeprecationWarning: Right multiplication will be prohibited in version 3.0
    x, _ = (np.arange(nx) + 0.5, np.zeros(nx) + 0.5) * transform
```

PR #57, opened 2019-11-01
add 'variable' and 'group' kwargs to 'rioxarray.open_rasterio()' to allow filtering of subdatasets

snowman2:
- [x] Tests added

### DIFF
--- a/rioxarray/_io.py
+++ b/rioxarray/_io.py
@@ -7,6 +7,7 @@ Source file: https://github.com/pydata/xarray/blob/1d7bcbdc75b6d556c04e2c7d7a042
 """
 
 import os
+import re
 import warnings
 from collections import OrderedDict
 from distutils.version import LooseVersion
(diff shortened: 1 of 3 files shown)


rioxarray/_io.py
```diff
@@ -194,14 +195,49 @@ def parsevec(s):
...
+ 
+    Returns
+    -------
+    re.SRE_Pattern: output of re.compile()
+    """
+    variable_query = r"\w+"
+    if variable_names is not None:
+        if not isinstance(variable_names, (tuple, list)):
+            variable_names = [variable_names]
+        variable_query = rf"(?:{'|'.join(variable_names)})"
```
justingruca: Are there any restrictions on the syntax of `variable_names`? Might want to run them through [`re.escape`](https://docs.python.org/3/library/re.html#re.escape).

rioxarray/_io.py
```diff
@@ -194,14 +195,49 @@ def parsevec(s):
...
+    if group_names is not None:
+        if not isinstance(group_names, (tuple, list)):
+            group_names = [group_names]
+        group_query = rf"(?:{'|'.join(group_names)})"
```
justingruca: Same with `group_names`.

rioxarray/_io.py
```diff
@@ -194,14 +195,49 @@ def parsevec(s):
     return parsed_tags
 
 
+def build_subdataset_filter(group_names, variable_names):
+    """
+    Examples:
+'HDF4_EOS:EOS_GRID:"./modis/MOD09GQ.A2017290.h11v04.006.NRT.hdf":MODIS_Grid_2D:sur_refl_b01_1'
```
justingruca: indent?

snowman2: Fixed that one.

rioxarray/_io.py
snowman2: Good suggestion. Added it in. Thanks!

PR #68, opened 2019-11-22
Write raster metadata

### LINKED ISSUE #64: Write out metadata to tif files on export
1. Preserve attributes in the tags
2. Band names: https://gis.stackexchange.com/a/284242/144357

snowman2:
- [x] Closes #64
 - [x] Tests added

NOTE: removed writing creation_date on raster operations. Not necessary and probably unwanted.


rioxarray/rioxarray.py
```diff
@@ -1177,6 +1177,27 @@ def to_raster(
             ),
             **out_profile,
         ) as dst:
+            # write metadata to raster
```
snowman2: Hmmm, this method is getting a bit too long.

rioxarray/rioxarray.py
```diff
@@ -1177,6 +1177,27 @@ def to_raster(
...
+            tags = {} if tags is None else tags
+            dst.update_tags(**self._obj.attrs, **tags)
+
+            # write band name information
+            long_name = self._obj.attrs.get("long_name")
+            if isinstance(long_name, (tuple, list)):
```
snowman2: Need test for this case.

snowman2: Added

rioxarray/rioxarray.py
snowman2: Moved to separate function

PR #79, opened 2020-01-20
 Add support for writing scales & offsets to raster; Don't write standard raster metadata to raster tags

### LINKED ISSUE #78: Saving a RasterArray with a float nodata value encodes it as a string and transform information is corrupted
I'm on the most recent version of rioxarray, .18

The data file I'm using, "NEON-DS-Airborne-Remote-Sensing/HARV/DTM/HARV_dtmCrop.tif" can be downloaded here: https://ndownloader.figshare.com/articles/2009586/versions/10

```python
DTM_UTM = xarray.open_rasterio("data/NEON-DS-Airborne-Remote-Sensing/HARV/DTM/HARV_dtmCrop.tif")

from rasterio.crs import CRS
crs = CRS.from_epsg(4326)

reprojected = DTM_UTM.rio.reproject(dst_crs=crs)
# this is a float
reprojected.rio.nodata
reprojected_path = "data/NEON-DS-Airborne-Remote-Sensing/HARV/DTM/HARV_dtmCrop_WGS84.tif"
reprojected.rio.to_raster(reprojected_path)

DTM_WGS84 = xarray.open_rasterio(reprojected_path)
DTM_WGS84.rio.nodata # this is now a string because it was saved to a raster but it should be a float or int
```
```
'-9999.0'
```
Is this fixable? I tried converting the nodata value to an int first before saving but the issue still came up. I see that there is also a `nodatavals` attribute in my dataset that actually contains a float in a tuple, so I can use that, but I'm curious if it is possible to handle the `nodata` value issue since I'm designing a geospatial python lesson that uses rioxarray. Thanks for the great library!

snowman2:
- [x] Closes #78
 - [x] Tests added

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -1,6 +1,11 @@
 History
 =======
 
+0.0.19
+-------
+- Add support for writing scales & offsets to raster (pull #79)
+- Don't write standard raster metadata to raster tags (issue #78)
+
 0.0.18
 ------
 - Fixed windowed writing to require tiled output raster (pull #66)
--- a/rioxarray/rioxarray.py
+++ b/rioxarray/rioxarray.py
@@ -209,9 +209,32 @@ def _write_metatata_to_raster(raster_handle, xarray_dataset, tags):
     """
     Write the metadata stored in the xarray object to raster metadata
     """
-    # write metadata to raster
-    tags = {} if tags is None else tags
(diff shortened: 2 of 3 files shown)


snowman2:
Failures are unrelated to this PR - looks like GDAL 2.4.4 broke some things.

PR #116, opened 2020-06-10
Allow destination shape to be specified in reproject()

davidbrochart:
- [x] Tests added

### DIFF
--- a/rioxarray/rioxarray.py
+++ b/rioxarray/rioxarray.py
@@ -209,12 +209,16 @@ def _make_dst_affine(
     """Determine the affine of the new projected `xarray.DataArray`"""
     src_bounds = src_data_array.rio.bounds()
     src_width, src_height = src_data_array.rio.shape
-    resolution_or_width_height = {}
-    if dst_resolution is not None:
-        resolution_or_width_height["resolution"] = dst_resolution
-    if dst_shape is not None:
-        resolution_or_width_height["dst_height"] = dst_shape[0]
-        resolution_or_width_height["dst_width"] = dst_shape[1]
+    dst_height, dst_width = dst_shape if dst_shape is not None else (None, None)
+    resolution_or_width_height = {
+        k: v
+        for k, v in [
+            ("resolution", dst_resolution),
+            ("dst_height", dst_height),
+            ("dst_width", dst_width),
+        ]
+        if v is not None
+    }
     dst_affine, dst_width, dst_height = rasterio.warp.calculate_default_transform(
         src_crs,
         dst_crs,
--- a/rioxarray/rioxarray.py
+++ b/rioxarray/rioxarray.py
@@ -211,13 +211,17 @@ def _make_dst_affine(
     src_width, src_height = src_data_array.rio.shape
     resolution_or_width_height = {}
     if dst_resolution is not None:
-        resolution_or_width_height['resolution'] = dst_resolution
+        resolution_or_width_height["resolution"] = dst_resolution
     if dst_shape is not None:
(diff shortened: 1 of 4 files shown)


rioxarray/rioxarray.py
```diff
@@ -871,6 +885,8 @@ def reproject(
         resolution: float or tuple(float, float), optional
             Size of a destination pixel in destination projection units
             (e.g. degrees or metres).
+        shape: tuple(int, int), optional
+            Shape of the destination in pixels. Cannot be used together with resolution.
```
snowman2, line 889: Mind adding a note that the order in the tuple is (dst_height, dst_width)?

snowman2:
@davidbrochart, this looks great, thanks :+1:. Just a minor comment above and I think that it would be good to add another test that ensures the output shape is expected when using the `shape` argument. Would you mind adding that in?

snowman2:
Also, just thought of this - mind adding `.. versionadded:: 0.0.27 shape` to above the parameters section in the docstrings of both `reproject` methods?

rioxarray/rioxarray.py
davidbrochart: :+1:

rioxarray/rioxarray.py
```diff
@@ -203,12 +203,29 @@ def _make_coords(src_data_array, dst_affine, dst_width, dst_height, dst_crs):
...
 
-def _make_dst_affine(src_data_array, src_crs, dst_crs, dst_resolution=None):
+def _make_dst_affine(
+    src_data_array, src_crs, dst_crs, dst_resolution=None, dst_shape=None
+):
     """Determine the affine of the new projected `xarray.DataArray`"""
     src_bounds = src_data_array.rio.bounds()
     src_width, src_height = src_data_array.rio.shape
+    dst_height, dst_width = dst_shape if dst_shape is not None else (None, None)
+    resolution_or_width_height = {
```
davidbrochart, line 213: Should `shape` be `(dst_width, dst_height)`, to conform to the source shape?

snowman2: If you explicitly pass in `dst_width` `dst_height`, rasterio will do the validation for you:
https://github.com/mapbox/rasterio/blob/ce00d482faa24068f41164b79901eeec31777263/rasterio/warp.py#L457-L467

snowman2: > Should shape be (dst_width, dst_height), to conform to the source shape?

That is actually a bug I think - those should be the other way around.

snowman2: > should we explicitly pass dst_width and dst_height?

I am fine either way. I like `shape` but `dst_width` and `dst_height` are used by `rasterio` and make the code simpler as you just pass in the values.

davidbrochart: I was also thinking that rioxarray offers a somewhat higher-level API than rasterio, and more similar to NumPy. That's why I thought that `(dst_height, dst_width)` is really the shape of the array. So I'm fine with keeping things like that, if you tell me that the source shape should also be `(src_height, src_width)`.

snowman2: Sounds good, then I think it should be good to go as is.

snowman2:
@davidbrochart did you want to squash your commits with me doing a merge commit or would you rather I just do a squash-and-merge?

davidbrochart:
Yes, please go ahead and do a squash-and-merge.

snowman2:
Thanks @davidbrochart for this addition :+1:

davidbrochart:
Thanks a lot for this awesome library!

PR #138, opened 2020-06-20
ENH: Add pad_box()

### LINKED ISSUE #137: Make clip_box extend array
Today `clip_box` has an effect only when the array is shrunk. It would be nice if it could extend the array when the bounding box is bigger than the original one, maybe using xarray's `.pad()`?

davidbrochart:
- [x] Closes #137
 - [x] Tests added


rioxarray/rioxarray.py
```diff
@@ -1051,8 +1051,81 @@ def slice_xy(self, minx, miny, maxx, maxy):
...
+        superset = self._obj.pad(
+            pad_width={
+                self.x_dim: (x_before, x_after),
+                self.y_dim: (y_before, y_after),
+            },
+            constant_values=np.nan,
+        ).rio.set_spatial_dims(x_dim=self.x_dim, y_dim=self.y_dim, inplace=True)
+        superset[self.x_dim] = x_coord
+        superset[self.y_dim] = y_coord
+        superset.attrs["transform"] = tuple(self.transform(recalc=True))
```
snowman2, line 1115:
```suggestion
        superset.attrs["transform"] = tuple(superset.rio.transform(recalc=True))
```

rioxarray/rioxarray.py
```diff
@@ -1051,8 +1051,81 @@ def slice_xy(self, minx, miny, maxx, maxy):
         subset.attrs["transform"] = tuple(self.transform(recalc=True))
         return subset
 
-    def clip_box(self, minx, miny, maxx, maxy, auto_expand=False, auto_expand_limit=3):
-        """Clip the :class:`xarray.DataArray` by a bounding box.
+    def pad_xy(self, minx, miny, maxx, maxy, resolution_x, resolution_y):
```
snowman2, line 1054: I am wondering if resolution_x and resolution_y need to be required here. My thoughts are that they could be pulled in from the dataset the same way it was done in `clip_box`.

rioxarray/rioxarray.py
```diff
@@ -1051,8 +1051,81 @@ def slice_xy(self, minx, miny, maxx, maxy):
...
+            x_after = len(new_x_coord) - len(x_coord)
+            x_coord = new_x_coord
+            right = x_coord[-1]
+
+        superset = self._obj.pad(
+            pad_width={
+                self.x_dim: (x_before, x_after),
+                self.y_dim: (y_before, y_after),
+            },
+            constant_values=np.nan,
```
snowman2, line 1111: I haven't tested this out yet - does this fill the generated area with `NaN`?

rioxarray/rioxarray.py
davidbrochart: :+1:

rioxarray/rioxarray.py
davidbrochart: Yes, you can see it in the `clip_box.ipynb` notebook.

davidbrochart:
Actually there is a bug, the padding doesn't take into account the previous clipping. So maybe `.clip_box()` should be left as it is, and we should have a new `.pad_box()` method? Things like clipping on the right and padding on the left would be achieved by calling `.clip_box()` then `.pad_box()`.

rioxarray/rioxarray.py
```diff
@@ -1051,6 +1051,101 @@ def slice_xy(self, minx, miny, maxx, maxy):
...
+        ).rio.set_spatial_dims(x_dim=self.x_dim, y_dim=self.y_dim, inplace=True)
+        superset[self.x_dim] = x_coord
+        superset[self.y_dim] = y_coord
+        superset.attrs["transform"] = tuple(superset.rio.transform(recalc=True))
+        return superset
+
+    def pad_box(self, minx, miny, maxx, maxy):
+        """Pad the :class:`xarray.DataArray` to a bounding box
+
```
snowman2, line 1117: Mind adding `.. versionadded:: 0.0.29`?

rioxarray/rioxarray.py
```diff
@@ -1051,6 +1051,101 @@ def slice_xy(self, minx, miny, maxx, maxy):
         subset.attrs["transform"] = tuple(self.transform(recalc=True))
         return subset
 
+    def pad_xy(self, minx, miny, maxx, maxy):
+        """Pad the array to x,y bounds.
+
```
snowman2, line 1056: Mind adding `.. versionadded:: 0.0.29`?

rioxarray/rioxarray.py
```diff
@@ -1051,6 +1051,105 @@ def slice_xy(self, minx, miny, maxx, maxy):
...
+            x_after = len(new_x_coord) - len(x_coord)
+            x_coord = new_x_coord
+            right = x_coord[-1]
+
+        superset = self._obj.pad(
+            pad_width={
+                self.x_dim: (x_before, x_after),
+                self.y_dim: (y_before, y_after),
+            },
+            constant_values=np.nan,
```
snowman2, line 1110: Thoughts on attempting to get the nodata value to use for `constant_values`?

Something like:
```python
nodata = self.encoded_nodata if self.encoded_nodata is not None else self.nodata
if nodata is None:
    nodata = np.nan
```

davidbrochart: Sounds good, or even passing `constant_values` to `.pad_box(minx, miny, maxx, maxy, constant_values=None)` (and `pad_xy`), then in `pad_xy`:
```python
if constant_values is None:
    constant_values = self.encoded_nodata if self.encoded_nodata is not None else self.nodata
if constant_values is None:
    constant_values = np.nan
```
i.e. if the user knows what he wants, he should pass it explicitly, and otherwise we use the value most likely to represent no data. What do you think?

snowman2: That makes sense to me.

snowman2:
@davidbrochart this looks great, thanks!

PR #140, opened 2020-06-23
BUG: Fix assigning fill value in rio.pad_box

snowman2:
Discovered this while debugging something else. 

I was incorrect about the `encoded_nodata` as the data in the array will be `np.nan` when `encoded_nodata` is set. In this case, it would be better to leave it as `np.nan`.

cc: @davidbrochart 

 - [x] Tests updated


davidbrochart:
:+1:

PR #141, opened 2020-06-23
ENH: Add write_transform to store cache in GDAL location

### LINKED ISSUE #129: Store transform where GDAL will find it
https://gdal.org/drivers/raster/netcdf.html#georeference

```
 NetCDF driver will try to read the following metadata directly and set up georeferencing.
    spatial_ref (Well Known Text)
    GeoTransform (GeoTransform array)
```
It is probably worth doing for the case where `parse_coords=False` when reading a file and then it is written to netCDF. Also, I believe this is another part that goes along with #81

Currently, it is just stored in the `transform` attribute.

### LINKED ISSUE #139: Store transform
As described [here](https://corteva.github.io/rioxarray/stable/examples/manage_information_loss.html), information can be lost when doing xarray operations. Similar to the `write_crs` method, should we have a `write_transform` method that would ensure the transform gets propagated across operations? It also makes sense because some rioxarray operations, like `clip_box`, could update the transform.

snowman2:
- [x] Closes #129
-  [x] Closes #139
 - [x] Tests added

- Noticed that the `sel()` operation returned an array that way sharing the coordinates of the parent for some reason ... so added `copy()` in `slice_xy()`.
- Noticed some test were using fixture instead of parametrize & weren't being run

### DIFF
--- a/rioxarray/rioxarray.py
+++ b/rioxarray/rioxarray.py
@@ -78,7 +78,7 @@ def _get_grid_map_name(src_data_array):
         return DEFAULT_GRID_MAP
 
 
-def _generate_attrs(src_data_array, dst_affine, dst_nodata):
+def _generate_attrs(src_data_array, dst_nodata):
     # add original attributes
     new_attrs = copy.deepcopy(src_data_array.attrs)
     # remove all nodata information
@@ -95,7 +95,6 @@ def _generate_attrs(src_data_array, dst_affine, dst_nodata):
         new_attrs["_FillValue"] = fill_value
 
     # add raster spatial information
-    new_attrs["transform"] = tuple(dst_affine)[:6]
     new_attrs["grid_mapping"] = _get_grid_map_name(src_data_array)
 
     return new_attrs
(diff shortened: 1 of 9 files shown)


rioxarray/rioxarray.py
```diff
@@ -859,9 +917,12 @@ def transform_bounds(self, dst_crs, densify_pts=21, recalc=False):
         )
 
     def transform(self, recalc=False):
-        """Determine the affine of the `xarray.DataArray`"""
-        src_left, _, _, src_top = self.bounds(recalc=recalc)
-        src_resolution_x, src_resolution_y = self.resolution(recalc=recalc)
+        """affine.Afffine: The affine of the `xarray.DataArray`"""
```
jgruca, line 920: Is this docstring correct?

snowman2: Good catch - needed some updates.

snowman2:
Appveyor failure seems to be unrelated. I am guessing to an update in scipy.

snowman2:
Windows has scipy 1.5 & linux/osx has 1.4

PR #144, opened 2020-06-24
TST: Update interpolate tests for scipy 1.5

snowman2:


snowman2:
Well, that's unfortunate. It passes on Linux.

PR #210, opened 2021-01-19
ENH: Add support for using dask in 'rio.to_raster'

### LINKED ISSUE #9: Use windowed writing in to_raster
From: https://github.com/corteva/rioxarray/pull/8

> The only other helpful long term solution would be to use windowed writing like we do in trollimage to achieve better performance with dask (saving chunk by chunk), but this has its own issues we've discovered with GDAL's in-memory caching and making large output files if you don't have a lot of memory available.

ping @djhoese

snowman2:
- [x] Closes #9
 - [x] Tests added

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,6 +68,9 @@ The *open_rasterio* functionality was adopted from https://github.com/pydata/xar
   - `xarray is licensed <https://github.com/pydata/xarray/blob/1d7bcbdc75b6d556c04e2c7d7a042e4379e15303/LICENSE>`_ under the Apache License, Version 2.0.
     The xarray license is included as `LICENSE_xarray <https://github.com/corteva/rioxarray/blob/master/LICENSE_xarray>`_.
 
+RasterioWriter dask write functionality was adopted from https://github.com/dymaxionlabs/dask-rasterio
+  - Source file: `write.py <https://github.com/dymaxionlabs/dask-rasterio/blob/8dd7fdece7ad094a41908c0ae6b4fe6ca49cf5e1/dask_rasterio/write.py>`_
+
 
 This package was originally templated with with Cookiecutter_.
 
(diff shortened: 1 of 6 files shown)


justingruca (approved):

djhoese:
I try my best to keep up with rioxarray, but you work too fast. Quick question on this very neat feature: any idea if it works with distributed schedulers (multi-process or multi-node)?

snowman2:
> any idea if it works with distributed schedulers (multi-process or multi-node)?

I tried out on a cluster and it seems to fail due to these issues:
- https://github.com/dymaxionlabs/dask-rasterio/issues/3
- https://github.com/dask/dask/issues/3255

I see you have been searching for a solution to this @djhoese. I will need to look into it.

djhoese:
Yes. In general I think it is a limitation of the underlying geotiff format and the C/C++ library (not multiprocess safe).

snowman2:
@djhoese I found a way for it to work based on https://github.com/dymaxionlabs/dask-rasterio/issues/3#issuecomment-514781825.

jgruca (approved):

djhoese:
Does this PR allow me to get a Delayed object back and compute/write the result later? Glad you got something that works. I'll have to convert some of Satpy's stuff to use rioxarray directly now.

snowman2:
> Does this PR allow me to get a Delayed object back and compute/write the result later?

I don't think it will give you a delayed object back currently. But, feel free to open up a new issue for this functionality.

djhoese:
I'll try to find time to do it later today. That would be the only missing feature needed to cleaning swap our current rasterio geotiff writer for rioxarray in Satpy. We use it so we can say multiple RGB composites at once that may be made up of only 4 or 5 original bands. By delaying the computation we can let dask "cache" the input bands so it only computes them once and reuses them for each geotiff that needs to be saved.

Issue #214, opened 2021-01-26
Issue with locking when reading from multiple threads with Dask


TomAugspurger:
Currently, rioxarray.open_rasterio builds a `DataArray` with a `_file_obj` that's an xarray `CacheFileManager` that eventually calls `rasterio.open` on the user-provided input (typically a filename). When using rioxarray with `chunks=True` to do parallel reads from multiple threads this can cause workers to wait for a lock, since GDAL file objects are not safe to be read from multiple threads.

This snippet demonstrates the issue by creating and then reading a `(4, 2560, 2560)` TIF. It uses a SerializableLock subclass that adds a bit of logging when locks are requested and acquired.

```python
# file: rioxarray-locks.py
import rasterio
import rioxarray
import numpy as np
import logging
import threading
import time
import dask
from dask.utils import SerializableLock

logger = logging.getLogger("lock")
logger.setLevel(logging.INFO)
handler = logging.FileHandler("lock.log")
handler.setLevel(logging.INFO)
logger.addHandler(handler)


class InstrumentedLock(SerializableLock):
    """A lock that logs acquisition attempts and completions."""
    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        
    def __enter__(self, *args, **kwargs):
        ident = threading.get_ident()
        t0 = time.perf_counter()
        logger.info("GET lock %d from %d", id(self.lock), ident)
        super().__enter__()
        logger.info("GOT lock %d from %d waited %f", id(self.lock), ident, time.perf_counter() - t0)


if __name__ == "__main__":
    # Generate a tiff file. Has 4 bands that could be read in parallel.
    a = np.random.randint(0, 255, size=(4, 10 * 256, 10 * 256), dtype="uint8")
    transform = rasterio.Affine(30.0, 0.0, 358485.0, 0.0, -30.0, 4265115.0)
    count, width, height = a.shape

    with rasterio.open('example.tif', 'w',
                       width=width, height=height, count=count,
                       dtype=rasterio.uint8,
                       crs='+proj=latlong', transform=transform) as dst:
        for band in range(count):
            dst.write(a[band], band + 1)


    lock = InstrumentedLock()
    ds = rioxarray.open_rasterio("example.tif", chunks=True, lock=lock)

    logger.info("-------------- read meatadata ------------------")

    with dask.config.set(scheduler="threads"):
        ds.mean().compute()
```

Here's the output to `lock.log:

```
GET lock 139663598009152 from 139664510379840
[...]

PR #219, opened 2021-01-26
ENH: Add 'compute' keyword argument to 'to_raster'

### LINKED ISSUE #211: Add 'compute' keyword argument to dask rasterio writer
Thanks to #210 it is now possible to save dask arrays with rasterio. One feature I'd really like to see so I can completely replace my existing geotiff writer (in Satpy's dependent package trollimage) is the ability to return a dask Delayed object by passing `compute=False` to the `dask.array.store` function. See store's documentation here: https://docs.dask.org/en/latest/array-api.html#dask.array.store

Complications I see with this are that currently the `to_raster` method doesn't return anything. Another one is that I've had issues in the past with trollimage handling the delayed objects because it required opening the file object (the rasterio one) and then requiring the user to close it (return the open file object handle and the delayed object). My understanding of #210 is that you aren't opening the file until it is being written to so this may not be an issue.

djhoese:
See #211 for additional context. The summary is that dask's `store` function allows you to pass `compute=False`. Doing this returns a `Delayed` object which can be computed later. This is really useful if your input dask arrays are used in other outputs (very common in the Satpy library where one input raster may be used for different RGB channels in various composites).

At the time of writing this only implements the functionality and includes some docstrings. Some questions and issues I ran into that I'm looking for help on:

1. Where would you like tests added for this? I could add more parametrize cases to `test_to_raster` in `test_integration_rioxarray`.
2. This, as I feared in #211, doesn't seem to work on a distributed client (multiprocess). But this doesn't actually seem to work with the existing behavior (compute=True). See below.
3. Other feedback?

## Distributed Issues

Example:

```
import rioxarray
import dask.diagnostics
import dask.distributed

g = rioxarray.open_rasterio("any_geotiff.tif", chunks=1024)
client = dask.distributed.Client()
with dask.diagnostics.ProgressBar():
    delayed = g.rio.to_raster("/tmp/some_out.tif", windowed=True)

# or
with dask.diagnostics.ProgressBar():
    delayed = g.rio.to_raster("/tmp/some_out.tif", windowed=True, compute=False)
with dask.diagnostics.ProgressBar():
    delayed.compute()
```

```
~/miniconda3/envs/satpy_py37/lib/python3.7/site-packages/dask/array/core.py in load_store_chunk()
   3711     try:
   3712         if x is not None:
-> 3713             out[index] = np.asanyarray(x)
   3714         if return_stored and load_stored:
   3715             result = out[index]

~/repos/git/rioxarray/rioxarray/raster_writer.py in __setitem__()
    121
    122         with rasterio.open(self.raster_path, "r+") as rds:
--> 123             rds.write(item, window=Window(chx_off, chy_off, chx, chy), indexes=indexes)
    124
    125     def to_raster(self, xarray_dataarray, tags, windowed, compute, **kwargs):

rasterio/_io.pyx in rasterio._io.DatasetWriterBase.write()

RasterioIOError: Read or write failed. /tmp/c07.tif, band 1: IReadBlock failed at X offset 0, Y offset 1108: TIFFReadEncodedStrip() failed.
```

Ideas?

## Checklist

 - [ ] Closes #211
 - [ ] Tests added

### DIFF
--- a/rioxarray/raster_array.py
+++ b/rioxarray/raster_array.py
@@ -851,7 +851,7 @@ class RasterArray(XRasterBase):
         lock: boolean or Lock, optional
             Lock to use to write data using dask.
             If not supplied, it will use a single process for writing.
-        compute: bool
+        compute: bool, optional
             If True and data is a dask array, then compute and save
             the data immediately. If False, return a dask Delayed object.
             Call ".compute()" on the Delayed object to compute the result
--- a/rioxarray/raster_array.py
+++ b/rioxarray/raster_array.py
@@ -865,7 +865,7 @@ class RasterArray(XRasterBase):
         Returns
         -------
         :obj:`dask.Delayed`:
-            If the data array is a dask array, windowed is True, and compute
+            If the data array is a dask array and compute
             is True. Otherwise None is returned.
 
         """
--- a/rioxarray/raster_array.py
+++ b/rioxarray/raster_array.py
@@ -824,6 +824,7 @@ class RasterArray(XRasterBase):
         windowed=False,
         recalc_transform=True,
         lock=None,
+        compute=True,
         **profile_kwargs,
     ):
         """
@@ -850,10 +851,23 @@ class RasterArray(XRasterBase):
         lock: boolean or Lock, optional
             Lock to use to write data using dask.
             If not supplied, it will use a single process for writing.
+        compute: bool
+            If True and the data array is a dask array and windowed
+            is True, then compute and save the data immediately. If False,
+            return a dask Delayed object. Call ".compute()" on the Delayed
+            object to compute the result later. Call
+            ``dask.compute(delayed1, delayed2)`` to save multiple delayed
+            files at once. Default is True.
         **profile_kwargs
             Additional keyword arguments to pass into writing the raster. The
             nodata, transform, crs, count, width, and height attributes
             are ignored.
+
+        Returns
+        -------
+        :obj:`dask.Delayed`:
+            If the data array is a dask array, windowed is True, and compute
+            is True. Otherwise None is returned.
 
         """
         if driver is None and LooseVersion(rasterio.__version__) < LooseVersion("1.2"):
@@ -893,7 +907,7 @@ class RasterArray(XRasterBase):
             # converted right before writing.
             rio_nodata = _ensure_nodata_dtype(rio_nodata, dtype)
 
-        RasterioWriter(raster_path=raster_path).to_raster(
+        return RasterioWriter(raster_path=raster_path).to_raster(
             xarray_dataarray=self._obj,
             tags=tags,
             driver=driver,
@@ -906,5 +920,6 @@ class RasterArray(XRasterBase):
             nodata=rio_nodata,
             windowed=windowed,
             lock=lock,
+            compute=compute,
             **out_profile,
         )
--- a/rioxarray/raster_array.py
+++ b/rioxarray/raster_array.py
@@ -852,12 +852,11 @@ class RasterArray(XRasterBase):
             Lock to use to write data using dask.
             If not supplied, it will use a single process for writing.
         compute: bool
-            If True and the data array is a dask array and windowed
-            is True, then compute and save the data immediately. If False,
-            return a dask Delayed object. Call ".compute()" on the Delayed
-            object to compute the result later. Call
-            ``dask.compute(delayed1, delayed2)`` to save multiple delayed
-            files at once. Default is True.
+            If True and data is a dask array, then compute and save
+            the data immediately. If False, return a dask Delayed object.
+            Call ".compute()" on the Delayed object to compute the result
+            later. Call ``dask.compute(delayed1, delayed2)`` to save
+            multiple delayed files at once. Default is True.
         **profile_kwargs
             Additional keyword arguments to pass into writing the raster. The
             nodata, transform, crs, count, width, and height attributes
--- a/rioxarray/raster_dataset.py
+++ b/rioxarray/raster_dataset.py
@@ -362,12 +362,11 @@ class RasterDataset(XRasterBase):
             Lock to use to write data using dask.
             If not supplied, it will use a single process for writing.
         compute: bool
-            If True and the data array is a dask array and windowed
-            is True, then compute and save the data immediately. If False,
-            return a dask Delayed object. Call ".compute()" on the Delayed
-            object to compute the result later. Call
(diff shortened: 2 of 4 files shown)


djhoese:
Crap, I wanted this to be a draft PR. It isn't done. Can I change it?

djhoese:
Ah ok the error is due to reading not being dask friendly, right? Seems related to #214 maybe.

If I do a `g = g.persist()` before trying `to_raster` then it seems to work (it completes without error). However, the image is corrupt (missing tiles).

image

snowman2:
> Where would you like tests added for this?

That is a good question. I think it would be good to check the return. For clarity, I am thinking it would make sense to add a separate test. But, if it is a minor tweak to the existing test, then it would probably be fine to add to the parametrized test.

snowman2:
> I wanted this to be a draft PR. It isn't done. Can I change it?

That would be a nice feature. I haven't found it yet, but if I do I will let you know. But, don't worry, I won't merge it in until you give the go ahead.

snowman2:
> However, the image is corrupt (missing tiles).

That's not good. I bet it has to do with aquiring the lock before writing. I need to test this on a bigger raster.

snowman2:
I think that the implementation you have seems fine. Would you mind also updating the RasterDataset `to_raster` method as well?

djhoese:
From the SerializableLock docstring:

```
    This wraps a normal ``threading.Lock`` object and satisfies the same
    interface.  However, this lock can also be serialized and sent to different
    processes.  It will not block concurrent operations between processes (for
    this you should look at ``multiprocessing.Lock`` or ``locket.lock_file``
    but will consistently deserialize into the same lock.
```

I can't find an easy way to get multiprocessing lock to work.

rioxarray/raster_array.py
```diff
@@ -771,11 +772,24 @@ def to_raster(
             This only works if the output raster is tiled. As such, if you
             set this to True, the output raster will be tiled.
             Default is False.
+        compute: bool
+            If True and the data array is a dask array and windowed
```
snowman2, line 776: When you say windowed, did you mean `tiled=True`?

djhoese: My understanding from your docstring for `windowed` was that dask isn't used unless the data array is a dask array and windowed=True. Obviously this compute kwarg only comes into play if dask is used so that's what I'm trying to say here.

snowman2: Probably need to fix the wording there. I was trying to say that `windowed` is ignored for dask arrays.

djhoese: Ah ok, so dask.array.store is **always** used if the data is a dask array, right?

snowman2: Currently that is the case. I am working on an implementation that requires lock to be passed in. So, if they don't have a lock passed in, it does not use `dask.array.store`.

djhoese: Darn, I was hoping that using the default `lock=True` for the threaded cases (dask default) would be good enough. Maybe a special section in the documentation for doing multiprocess/cluster work. I guess your way is safer.

snowman2: > I was hoping that using the default lock=True for the threaded cases (dask default) would be good enough.

I run into this issue with the defaultr: https://github.com/dask/distributed/issues/780
Do you have a configuration that works with the default?

djhoese: If you don't create the `Client` object and don't have any other custom configuration then dask *should* should be using the threaded scheduler and the default of `lock=True` will use a thread-safe lock. That worked for me yesterday with this PR. The problem is when using `Client` to create a distributed scheduler (either a local one with multiple processes instead of threads or a multi-node one) then the thread-safe lock doesn't work since it isn't multi-process/node safe. That's what that issue is referring to.

snowman2: Ah, interesting. That is good to know.

djhoese:
@snowman2 I just merged master and forced push here. I'm noticing that none of the tests actually cover the dask writing. Is that intentional?

rioxarray/raster_dataset.py
```diff
@@ -360,11 +361,23 @@ def to_raster(
...
+            multiple delayed files at once. Default is True.
         **profile_kwargs
             Additional keyword arguments to pass into writing the raster. The
             nodata, transform, crs, count, width, and height attributes
             are ignored.
 
+        Returns
+        -------
+        :obj:`dask.Delayed`:
+            If the data array is a dask array, windowed is True, and compute
```
snowman2, line 378:
```suggestion
            If the data array is a dask array and compute
```

rioxarray/raster_array.py
```diff
@@ -850,11 +851,23 @@ def to_raster(
...
+            multiple delayed files at once. Default is True.
         **profile_kwargs
             Additional keyword arguments to pass into writing the raster. The
             nodata, transform, crs, count, width, and height attributes
             are ignored.
 
+        Returns
+        -------
+        :obj:`dask.Delayed`:
+            If the data array is a dask array, windowed is True, and compute
```
snowman2, line 868:
```suggestion
            If the data array is a dask array and compute
```

rioxarray/raster_array.py
```diff
@@ -850,11 +851,23 @@ def to_raster(
         lock: boolean or Lock, optional
             Lock to use to write data using dask.
             If not supplied, it will use a single process for writing.
+        compute: bool
```
snowman2, line 854:
```suggestion
        compute: bool, optional
```

rioxarray/raster_dataset.py
```diff
@@ -360,11 +361,23 @@ def to_raster(
         lock: boolean or Lock, optional
             Lock to use to write data using dask.
             If not supplied, it will use a single process for writing.
+        compute: bool
```
snowman2, line 364:
```suggestion
        compute: bool, optional
```

snowman2:
@djhoese, this looks great. Thanks for fixing the tests :+1:. Just a couple of tweaks to address.

justingruca (approved):

djhoese:
@snowman2 thanks for the review. All done now.

snowman2 (approved):

snowman2:
@djhoese this looks great. Would you like to squash you commits or do you mind if I squash and merge?

djhoese:
Feel free to squash and merge.

snowman2:
Thanks @djhoese :+1:

PR #223, opened 2021-01-27
ENH: Add lock kwarg to to_raster and add examples of using dask for file io

### LINKED ISSUE #220: BUG: Image missing tiles with dask write
The image is corrupt (missing tiles).

image

_Originally posted by @djhoese in https://github.com/corteva/rioxarray/issues/219#issuecomment-767867834_

snowman2:
- [x] Closes #220
 - [x] Tests added

cc: @djhoese

### DIFF
--- a/rioxarray/raster_array.py
+++ b/rioxarray/raster_array.py
@@ -756,10 +756,13 @@ class RasterArray(XRasterBase):
         tags=None,
         windowed=False,
         recalc_transform=True,
+        lock=None,
         **profile_kwargs,
     ):
         """
         Export the DataArray to a raster file.
+
+        ..versionadded:: 0.2 lock
 
         Parameters
         ----------
@@ -774,9 +777,12 @@ class RasterArray(XRasterBase):
             A dictionary of tags to write to the raster.
         windowed: bool, optional
             If True, it will write using the windows of the output raster.
-            This only works if the output raster is tiled. As such, if you
-            set this to True, the output raster will be tiled.
+            This is useful for loading data in chunks when writing. Does not
+            do anything when writing with dask.
             Default is False.
+        lock: boolean or Lock, optional
+            Lock to use to write data using dask.
+            If not supplied, it will use a single process for writing.
         **profile_kwargs
             Additional keyword arguments to pass into writing the raster. The
             nodata, transform, crs, count, width, and height attributes
@@ -832,5 +838,6 @@ class RasterArray(XRasterBase):
             transform=self.transform(recalc=recalc_transform),
             nodata=rio_nodata,
             windowed=windowed,
+            lock=lock,
             **out_profile,
         )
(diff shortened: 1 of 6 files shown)


justingruca (approved):

rioxarray/raster_array.py
```diff
@@ -774,9 +777,12 @@ def to_raster(
             A dictionary of tags to write to the raster.
         windowed: bool, optional
             If True, it will write using the windows of the output raster.
-            This only works if the output raster is tiled. As such, if you
-            set this to True, the output raster will be tiled.
+            This is useful for loading data in chunks when writing. Does not
+            do anything when writing with dask.
```
djhoese, line 781: Does this still "force" tiled output files?

snowman2: No. You need to use `tiled=True` for that.

djhoese: Ok so this is a change in how windowed operates?

snowman2: It was changed back [here](https://github.com/corteva/rioxarray/pull/134) and I missed the docstrings.

djhoese: Ooohhh. That explains it. So if I do `to_raster(..., tiled=True, blockxsize=512, blockysize=512)` it will store the dask arrays with that internal geotiff tile size, right (and not the dask chunk size)?

snowman2: That's correct. Currently not automatically tiling according to the dask chunk size at the moment. So it needs to be done by the user.

djhoese: Good. Not sure auto-tiling based on chunk size would be a good idea anyway. Just checking.

djhoese:
I just found out that this exists and could maybe simplify this:

https://github.com/dask/dask/blob/f41325512a4251dbaa7f7c781414d913066a6c34/dask/utils.py#L1017-L1028

I found that some of the dask functions actually use this and default to it.

snowman2:
@djhoese that looks pretty neat. However, that won't work for `to_raster`. The reason being that we need a global lock as only one process can write to the file at a time [ref](https://corteva.github.io/rioxarray/stable/examples/dask_read_write.html#Multiple-worker-example).

djhoese:
Isn't that how this `to_hdf` function is using it: https://docs.dask.org/en/latest/dataframe-api.html#dask.dataframe.DataFrame.to_hdf

snowman2:
> Isn't that how this to_hdf function is using it

For GeoTiff you need the global lock or you get the problems you found in #220

djhoese:
So "global" as in the global scope of the calling module, right? Couldn't users use this utility function so they don't have to decide what lock to choose?

snowman2:
> So "global" as in the global scope of the calling module, right? 

It means that the lock needs to be shared by all processes and threads.

> Couldn't users use this utility function so they don't have to decide what lock to choose?

I don't think so. Based on the dask issues I have seen, the SerializableLock only locks the threads within the process that contains. And the multiprocessing lock doesn't work for a cluster of machines. So, the dask Lock is the only option. But, I recommend trying it out and seeing what happens. You should be able to try it out now and see if it works or not.

PR #224, opened 2021-01-28
Lockless multi-threaded reads

### LINKED ISSUE #214: Issue with locking when reading from multiple threads with Dask

TomAugspurger:
- [x] Closes #214 
 - [x] Tests added


TomAugspurger:
Here's a link for the new notebook: https://nbviewer.jupyter.org/github/corteva/rioxarray/blob/73ea5e24d968202f48ca8812f5c00104f9139638/docs/examples/read-locks.ipynb

snowman2:
This is really nice :+1:. I like that there are different options available for the file manager.

docs/history.rst
```diff
@@ -9,6 +9,7 @@ Latest
 - ENH: Use the list version of `transform_geom` with rasterio 1.2+ (issue #180)
 - ENH: Support driver autodetection with rasterio 1.2+ (issue #180)
 - BUG: Allow `rio.write_crs` when spatial dimensions not found (pull #186)
+- ENH: Allow multithreaded, lockless reads with `rio.read_xarary` (issue #214)
```
justingruca, line 12: `s/xarary/xarray/`

snowman2 (approved):

docs/examples/read-locks.ipynb
```diff
@@ -0,0 +1,227 @@
...
+    "\n",
+    "If you're using `rioxarray` with [Dask](http://docs.dask.org/) through the `chunks` keyword, you can also specify the `lock=False` keyword to ensure that reading *and* operating on your data happen in parallel.\n",
+    "\n",
+    "## Scheduler Choice\n",
+    "\n",
+    "Dask has [several schedulers](https://docs.dask.org/en/latest/scheduling.html) which run computations in parallel. Which scheduler is best depends on a variety of factors, including whether your computation holds Python's Global Interpreter Lock, whether how much data needs to be moved around, and whether you need more than one machine's computational power. This section about read-locks only applies if you have more than one thread in a process. This will happen with Dask's [local threaded scheduler](https://docs.dask.org/en/latest/scheduling.html#local-threads) and it's [distributed scheduler](https://distributed.dask.org/en/latest/) when configured to use more than one thread per worker.\n",
```
justingruca, line 15: `s/it's/its/`

justingruca (commented):
Looks good, just a couple text changes.

TomAugspurger:
Whoops, thanks. Fixed those typos.

snowman2:
@TomAugspurger would you like to squash your commits or do you mind if I do a squash and merge?

TomAugspurger:
Squash and merge is fine by me.

snowman2:
Thanks @TomAugspurger :+1:

TomAugspurger:
Thanks!

Issue #189, opened 2020-11-16
reproject on larger dataset


joemcglinchy:
Hello, I'm trying to use the reprojection capability on a global raster dataset. I can read the data no problem, and it is of shape `(band: 1, y: 43200, x: 86400)`. The data is located here and is public, so the code below should load the data for anyone:  https://earthlab-jmcglinchy.s3-us-west-2.amazonaws.com/for_public/lc_mosaic_2003.tif 

it seems like the attempt for a numpy array of same dimensions is causing the memory error. 
#### Code Sample, a copy-pastable example if possible

A "Minimal, Complete and Verifiable Example" will make it much easier for maintainers to help you:
http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports

```python
import rioxarray as rx

lulc = rx.open_rasterio('/vsis3/earthlab-jmcglinchy/for_public/lc_mosaic_2003.tif')
lulc.rio.reproject("+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs")
```
#### Problem description

I get the following traceback:

```
C:\software\Anaconda3\envs\x-python\lib\site-packages\rioxarray\rioxarray.py:270: UserWarning: The nodata value (-3.4e+38) has been automatically changed to (-3.3999999521443642e+38) to match the dtype of the data.
  f"The nodata value ({original_nodata}) has been automatically "
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\software\Anaconda3\envs\x-python\lib\site-packages\rioxarray\rioxarray.py", line 1194, in reproject
    source=self._obj.values,
  File "C:\software\Anaconda3\envs\x-python\lib\site-packages\xarray\core\dataarray.py", line 569, in values
    return self.variable.values
  File "C:\software\Anaconda3\envs\x-python\lib\site-packages\xarray\core\variable.py", line 510, in values
    return _as_array_or_item(self._data)
  File "C:\software\Anaconda3\envs\x-python\lib\site-packages\xarray\core\variable.py", line 272, in _as_array_or_item
    data = np.asarray(data)
  File "C:\software\Anaconda3\envs\x-python\lib\site-packages\numpy\core\_asarray.py", line 83, in asarray
    return array(a, dtype, copy=False, order=order)
  File "C:\software\Anaconda3\envs\x-python\lib\site-packages\xarray\core\indexing.py", line 685, in __array__
    self._ensure_cached()
[...]

PR #227, opened 2021-01-29
DOC: Add example of re-projecting large dataset using WarpedVRT

snowman2:
- [x] Related to: #222 and #189


docs/examples/reproject.ipynb
```diff
@@ -1099,6 +1102,53 @@
...
     "xds_utm.rio.crs"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Reproject Large Rasters with Virtual Warping\n",
+    "\n",
+    "Using [WarpedVRT](https://rasterio.readthedocs.io/en/latest/topics/virtual-warping.html) enables re-projection from disk and reuces the amount\n",
```
justingruca, line 1112: `s/reuces/reduces/`

justingruca (approved):
Looks good, just one text change.

docs/examples/reproject.ipynb
snowman2: Good catch. Thanks!

PR #281, opened 2021-04-05
Add very minimal xarray plugin for `engine="rasterio"`

### LINKED ISSUE #197: ENH: Add xarray entrypoint
Relevant issues:
- https://github.com/pydata/xarray/pull/4577
- https://github.com/pydata/xarray/issues/4697

alexamici:
Use xarray plugin infrastructure to add support for:

```python
>>> import xarray as xr
>>> xr.open_dataset("myfile.jp2", engine="rasterio")
<xarray.Dataset>
Dimensions:      (x: 19087, y: 3932)
Coordinates:
  * band         (band) int64 1
  * y            (y) float64 ...
  * x            (x) float64 ...
    spatial_ref  int64 0
Data variables:
    band_data     (band, y, x) float32 ...

```

also add automatic engine selection for files with `.tif` and `.geotif` extensions.

```python
>>> xr.open_dataset("myfile.tif")
<xarray.Dataset>
Dimensions:      (x: 19087, y: 3932)
Coordinates:
  * band         (band) int64 1
  * y            (y) float64 ...
  * x            (x) float64 ...
    spatial_ref  int64 0
Data variables:
    band_data        (band, y, x) float32 ...

```

 - [x] Closes #197 
 - [x] Tests added

### DIFF
--- a/docs/getting_started/getting_started.rst
+++ b/docs/getting_started/getting_started.rst
@@ -26,7 +26,7 @@ Why use :func:`rioxarray.open_rasterio` instead of `xarray.open_rasterio`?
 5. It adds the coordinate axis CF metadata.
 6. It loads raster metadata into the attributes.
 
-Starting with xarray 0.18.0 rioxarray enables passing `engine="rasterio"` to all xarray open functions:
+rioxarray 0.4+ enables passing `engine="rasterio"` to ``xarray.open_dataset`` and ``xarray.open_mfdataset``:
 
 .. code-block:: python
 
--- a/docs/getting_started/getting_started.rst
+++ b/docs/getting_started/getting_started.rst
@@ -26,7 +26,7 @@ Why use :func:`rioxarray.open_rasterio` instead of `xarray.open_rasterio`?
 5. It adds the coordinate axis CF metadata.
 6. It loads raster metadata into the attributes.
 
-rioxarray 0.4+ enables passing `engine="rasterio"` to ``xarray.open_dataset`` and ``xarray.open_mfdataset``:
+rioxarray 0.4+ enables passing `engine="rasterio"` to ``xarray.open_dataset`` and ``xarray.open_mfdataset`` for xarray 0.18+:
 
 .. code-block:: python
 
--- a/docs/getting_started/getting_started.rst
+++ b/docs/getting_started/getting_started.rst
@@ -26,6 +26,12 @@ Why use :func:`rioxarray.open_rasterio` instead of `xarray.open_rasterio`?
 5. It adds the coordinate axis CF metadata.
 6. It loads raster metadata into the attributes.
 
+Starting with xarray 0.18.0 rioxarray enables passing `engine="rasterio"` to all xarray open functions:
+
+.. code-block:: python
+
+    ds = xr.open_dataset("my.tif", engine="rasterio")
+
 
 .. toctree::
    :maxdepth: 1
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -3,6 +3,9 @@ History
 
 Latest
 ------
+- DEP: Python 3.6+ (issue #215)
+- DEP: xarray 0.17+ (needed for issue #282)
+- REF: Store `grid_mapping` in `encoding` instead of `attrs` (issue #282)
 
 0.3.2
 -----
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -4,6 +4,10 @@ History
 Latest
 ------
 
+0.3.2
+-----
+- BUG: handle lock=True in open_rasterio (issue #273)
+
 0.3.1
 ------
 - BUG: Compatibility changes with xarray 0.17 (issue #254)
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -7,6 +7,7 @@ Latest
 0.3.2
 -----
 - BUG: handle lock=True in open_rasterio (issue #273)
+- ENH: enable `engine="rasterio"` via xarray backend API (issue #197 pull #281)
 
 0.3.1
 ------
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -6,11 +6,11 @@ Latest
 - DEP: Python 3.6+ (issue #215)
 - DEP: xarray 0.17+ (needed for issue #282)
 - REF: Store `grid_mapping` in `encoding` instead of `attrs` (issue #282)
+- ENH: enable `engine="rasterio"` via xarray backend API (issue #197 pull #281)
 
 0.3.2
 -----
 - BUG: handle lock=True in open_rasterio (issue #273)
-- ENH: enable `engine="rasterio"` via xarray backend API (issue #197 pull #281)
 
 0.3.1
 ------
--- a/rioxarray/xarray_plugin.py
+++ b/rioxarray/xarray_plugin.py
@@ -4,11 +4,14 @@ import xarray as xr
 
 from . import _io
 
+CAN_OPEN_EXT = {"geotif", "geotiff", "j2k", "jp2", "tif", "tiff", "vrt"}
+
 
 class GdalBackend(xr.backends.common.BackendEntrypoint):
     def open_dataset(self, filename_or_obj, drop_variables=None):
-        ds = _io.open_rasterio(filename_or_obj).to_dataset("band")
-        ds = ds.rename({idx: f"band{idx}" for idx in ds.data_vars})
+        ds = _io.open_rasterio(filename_or_obj)
+        if isinstance(ds, xr.DataArray):
+            ds = ds.to_dataset(name="band_data")
         ds = ds.reset_coords("spatial_ref")
         return ds
 
@@ -17,4 +20,4 @@ class GdalBackend(xr.backends.common.BackendEntrypoint):
             _, ext = os.path.splitext(filename_or_obj)
         except TypeError:
             return False
-        return ext in {".tif", ".geotif", ".tiff", ".geotiff"}
+        return ext[1:].lower() in CAN_OPEN_EXT
--- a/rioxarray/xarray_plugin.py
+++ b/rioxarray/xarray_plugin.py
@@ -21,6 +21,9 @@ CAN_OPEN_EXTS = {
 
 
 class RasterioBackend(xr.backends.common.BackendEntrypoint):
+    """
+    .. versionadded:: 0.4
+    """
     def open_dataset(
         self,
         filename_or_obj,
--- a/rioxarray/xarray_plugin.py
+++ b/rioxarray/xarray_plugin.py
@@ -1,0 +1,18 @@
+import os.path
+
+import xarray as xr
+
+from . import _io
+
+
+class GdalBackend(xr.backends.common.BackendEntrypoint):
+    def open_dataset(self, filename_or_obj, drop_variables=None):
+        ds = _io.open_rasterio(filename_or_obj).to_dataset("band")
+        return ds.rename({idx: f"band{idx}" for idx in ds.data_vars})
+
+    def guess_can_open(self, filename_or_obj):
+        try:
+            _, ext = os.path.splitext(filename_or_obj)
+        except TypeError:
+            return False
+        return ext in {".tif", ".geotif", ".tiff", ".geotiff"}
--- a/rioxarray/xarray_plugin.py
+++ b/rioxarray/xarray_plugin.py
@@ -36,8 +36,10 @@ class RasterioBackend(xr.backends.common.BackendEntrypoint):
         variable=None,
         group=None,
         default_name="band_data",
-        open_kwargs={},
+        open_kwargs=None,
     ):
+        if open_kwargs is None:
+            open_kwargs = {}
         ds = _io.open_rasterio(
             filename_or_obj,
             mask_and_scale=mask_and_scale,
--- a/rioxarray/xarray_plugin.py
+++ b/rioxarray/xarray_plugin.py
@@ -33,7 +33,7 @@ class GdalBackend(xr.backends.common.BackendEntrypoint):
         )
         if isinstance(ds, xr.DataArray):
             ds = ds.to_dataset()
-        if decode_coords is True or decode_coords == "coordinate":
+        if decode_coords is True or decode_coords == "coordinates":
             ds = ds.reset_coords("spatial_ref")
         return ds
 
--- a/rioxarray/xarray_plugin.py
+++ b/rioxarray/xarray_plugin.py
@@ -8,10 +8,30 @@ CAN_OPEN_EXT = {"geotif", "geotiff", "j2k", "jp2", "tif", "tiff", "vrt"}
 
 
 class GdalBackend(xr.backends.common.BackendEntrypoint):
-    def open_dataset(self, filename_or_obj, drop_variables=None):
-        ds = _io.open_rasterio(filename_or_obj)
+    def open_dataset(
+        self,
+        filename_or_obj,
+        drop_variables=None,
+        mask_and_scale=False,
+        parse_coordinates=None,
+        lock=None,
+        masked=False,
+        variable=None,
+        group=None,
+        default_name="band_data",
+    ):
+        ds = _io.open_rasterio(
+            filename_or_obj,
+            mask_and_scale=mask_and_scale,
+            parse_coordinates=parse_coordinates,
+            lock=lock,
+            masked=masked,
+            variable=variable,
+            group=group,
+            default_name=default_name,
+        )
         if isinstance(ds, xr.DataArray):
-            ds = ds.to_dataset(name="band_data")
+            ds = ds.to_dataset()
         ds = ds.reset_coords("spatial_ref")
         return ds
 
--- a/rioxarray/xarray_plugin.py
+++ b/rioxarray/xarray_plugin.py
@@ -36,6 +36,7 @@ class RasterioBackend(xr.backends.common.BackendEntrypoint):
         variable=None,
         group=None,
         default_name="band_data",
+        open_kwargs={},
     ):
         ds = _io.open_rasterio(
             filename_or_obj,
@@ -46,6 +47,7 @@ class RasterioBackend(xr.backends.common.BackendEntrypoint):
             variable=variable,
             group=group,
             default_name=default_name,
+            **open_kwargs,
         )
         if isinstance(ds, xr.DataArray):
             ds = ds.to_dataset()
--- a/rioxarray/xarray_plugin.py
+++ b/rioxarray/xarray_plugin.py
@@ -4,15 +4,14 @@ import xarray as xr
 
 from . import _io
 
-CAN_OPEN_EXT = {"geotif", "geotiff", "j2k", "jp2", "tif", "tiff", "vrt"}
+CAN_OPEN_EXTS = {"geotif", "geotiff", "j2k", "jp2", "tif", "tiff", "vrt"}
 
 
-class GdalBackend(xr.backends.common.BackendEntrypoint):
+class RasterioBackend(xr.backends.common.BackendEntrypoint):
     def open_dataset(
         self,
         filename_or_obj,
         drop_variables=None,
-        decode_coords=True,
         mask_and_scale=False,
         parse_coordinates=None,
         lock=None,
@@ -33,8 +32,6 @@ class GdalBackend(xr.backends.common.BackendEntrypoint):
         )
         if isinstance(ds, xr.DataArray):
             ds = ds.to_dataset()
-        if decode_coords is True or decode_coords == "coordinates":
-            ds = ds.reset_coords("spatial_ref")
         return ds
 
     def guess_can_open(self, filename_or_obj):
@@ -42,4 +39,4 @@ class GdalBackend(xr.backends.common.BackendEntrypoint):
             _, ext = os.path.splitext(filename_or_obj)
         except TypeError:
             return False
-        return ext[1:].lower() in CAN_OPEN_EXT
+        return ext[1:].lower() in CAN_OPEN_EXTS
--- a/rioxarray/xarray_plugin.py
+++ b/rioxarray/xarray_plugin.py
@@ -12,6 +12,7 @@ class GdalBackend(xr.backends.common.BackendEntrypoint):
         self,
         filename_or_obj,
         drop_variables=None,
+        decode_coords=True,
         mask_and_scale=False,
         parse_coordinates=None,
         lock=None,
@@ -32,7 +33,8 @@ class GdalBackend(xr.backends.common.BackendEntrypoint):
         )
         if isinstance(ds, xr.DataArray):
             ds = ds.to_dataset()
-        ds = ds.reset_coords("spatial_ref")
+        if decode_coords is True or decode_coords=="coordinate":
+            ds = ds.reset_coords("spatial_ref")
         return ds
 
     def guess_can_open(self, filename_or_obj):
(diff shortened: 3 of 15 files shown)


rioxarray/xarray_plugin.py
```diff
@@ -0,0 +1,20 @@
...
+        ds = ds.rename({idx: f"band{idx}" for idx in ds.data_vars})
+        ds = ds.reset_coords("spatial_ref")
+        return ds
+
+    def guess_can_open(self, filename_or_obj):
+        try:
+            _, ext = os.path.splitext(filename_or_obj)
+        except TypeError:
+            return False
+        return ext in {".tif", ".geotif", ".tiff", ".geotiff"}
```
snowman2, line 20: What is `guess_can_open` used for? Is this in the scenario that `engine` isn't provided?

snowman2:
> Use xarray plugin infrastructure to add support for:
> 
> ```python
> >>> import xarray as xr
> >>> xr.open_dataset("myfile.jp2", engine="gdal")
> <xarray.Dataset>
> Dimensions:      (x: 19087, y: 3932)
> Coordinates:
>   * y            (y) float64 ...
>   * x            (x) float64 ...
>     spatial_ref  int64 0
> Data variables:
>     band1        (y, x) float32 ...
> Attributes:
>     scale_factor:              1.0
>     add_offset:                0.0
>     grid_mapping:              spatial_ref
> ```
> 
> also add automatic engine selection for files with `.tif` and `.geotif` extensinos.
> 
> ```python
> >>> xr.open_dataset("mygile.tif")
> <xarray.Dataset>
> Dimensions:      (x: 19087, y: 3932)
> Coordinates:
>   * y            (y) float64 ...
>   * x            (x) float64 ...
>     spatial_ref  int64 0
> Data variables:
>     band1        (y, x) float32 ...
> Attributes:
>     scale_factor:              1.0
>     add_offset:                0.0
>     grid_mapping:              spatial_ref
> ```

One thing I am noticing in these examples is that the attributes for the `band` data variables are on the `Dataset`. If you go this route, it would be important to make sure that those attributes are pushed to the `DataArray`.

rioxarray/xarray_plugin.py
snowman2: > also add automatic engine selection for files with .tif and .geotif extensinos.

Just seeing this. This sounds like it is the case.

snowman2, line 20: Just had this thought. With rasterio 1.2+ you can do:
```python
>>> from rasterio.drivers import raster_driver_extensions
>>> list(raster_driver_extensions())
['vrt', 'tif', 'tiff', 'ntf', 'img', 'asc', 'dt0', 'dt1', 'dt2', 'png', 'jpg', 'jpeg', 'gif', 'xpm', 'bmp', 'pix', 'map', 'mpr', 'mpl', 'rgb', 'hgt', 'ter', 'nc', 'lbl', 'cub', 'xml', 'ers', 'jp2', 'j2k', 'grb', 'grb2', 'grib2', 'rsw', 'rst', 'grd', 'rda', 'kml', 'kmz', 'webp', 'pdf', 'sqlite', 'mbtiles', 'cal', 'ct1', 'mrf', 'pgm', 'ppm', 'pnm', 'hdr', 'bt', 'lcp', 'gtx', 'gsb', 'gvb', 'kro', 'byn', 'err', 'dem', 'kea', 'bag', 'gen', 'blx', 'sdat', 'sg-grd-z', 'xyz', 'hf2', 'dat', 'sigdem', 'gpkg', 'bil']
```

alexamici: That's a long list! 😆 

We need to be careful as we don't want to override netCDF, GRIB or other data formats best accessed with a dedicated backend, so I would still keep a curated list of extensions that `rioxarray` offers to open without passing the `engine`.

alexamici:
> One thing I am noticing in these examples is that the attributes for the `band` data variables are on the `Dataset`. If you go this route, it would be important to make sure that those attributes are pushed to the `DataArray`.

You are right, I used `.to_netcdf` and started looking for fix-ups, but I missed the fact that array attributes where promoted to global attributes.

Anyway the whole PR is work in progress. If you like the idea we can discuss the direction to go here, or in the main issue.

rioxarray/xarray_plugin.py
```diff
@@ -0,0 +1,43 @@
...
+            parse_coordinates=parse_coordinates,
+            lock=lock,
+            masked=masked,
+            variable=variable,
+            group=group,
+            default_name=default_name,
+        )
+        if isinstance(ds, xr.DataArray):
+            ds = ds.to_dataset()
+        ds = ds.reset_coords("spatial_ref")
```
snowman2, line 35: This should be removed. `spatial_ref` should be a coordinate.

rioxarray/xarray_plugin.py
snowman2: I am not familiar with how `xarray` selects the driver they want. Is there a way for `xarray` to set the preferred order to try to open the dataset?

rioxarray/xarray_plugin.py
alexamici: I'm trying to follow the CF Conventions where `grid_mapping` refers to a non-coordinate variable:

https://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#grid-mappings-and-projections

and the `coordinates` attribute is intended to be used only for variable containing coordinate values, not the CRS.

I know it is annoying when you work with `DataArray`s that the `spatial_ref` variable is left in the `Dataset`, but that is the specification that most CF-aware tools expect.

By removing the `reset_coords` and saving with `to_netcdf` I obtain a netCDF with:

```
$ ncdump -h test.nc
netcdf test {
dimensions:
...
	short band_data(band, y, x) ;
		band_data:_FillValue = -9999s ;
		band_data:scale_factor = 1. ;
		band_data:add_offset = 0. ;
		band_data:grid_mapping = "spatial_ref" ;
		band_data:coordinates = "spatial_ref" ;
```

Where the `band_data:coordinates = "spatial_ref" ;` looks weird.

What is the reason to keep it in the `.coords`?

rioxarray/xarray_plugin.py
alexamici: There is no such feature at the moment, the order of testing is not specified and the first plugin that matches is selected.

Anyway the `guess_can_open` is only used when `engine` is not passed explicitly, so it is a convenience feature aimed at non-expert users.

I think it is safe to add to `CAN_OPEN_EXT` the extensions of GIS file formats, but not more scientific formats that have, ora are likely to have a dedicated plugin.

rioxarray/xarray_plugin.py
snowman2: Thanks for being thorough in your investigation :+1:

> What is the reason to keep it in the .coords?

The main reason is to preserve the CRS information in as non-lossy mechanism as possible that works with GDAL when you do `to_netcdf`. Without it, you cannot store the CRS on a standalone `DataArray` (the default return type with `open_rasterio`).

> that is the specification that most CF-aware tools expect

It works as a coordinate variable with GDAL (the main netCDF CF reading tool rioxarray supports) and ArcMap. Which CF tools does this not work with?

> I'm trying to follow the CF Conventions where grid_mapping refers to a non-coordinate variable

Might be worth raising an issue here for clarification: https://github.com/cf-convention/cf-conventions

alexamici: GDAL reads the CRS information in the `grid_mapping` variable just fine even if it is not a coordinate, I'm using this feature a lot and I just double checked that in fact works with the current implementation as well.

```
>>> ds = xr.open_dataset("test/test_data/input/cog.tif")
>>> ds
<xarray.Dataset>
Dimensions:      (band: 1, x: 500, y: 500)
Coordinates:
  * band         (band) int64 1
  * y            (y) float64 2.715e+06 2.714e+06 ... 2.565e+06 2.565e+06
  * x            (x) float64 1.635e+06 1.635e+06 ... 1.784e+06 1.784e+06
Data variables:
    spatial_ref  int64 ...
    band_data    (band, y, x) int16 ..
>>> ds.to_netcdf("test.nc")
>>> rds = rasterio.open("test.nc")
>>> rds.crs
CRS.from_wkt('PROJCS["Albers",GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378140,298.256999999996,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AUTHORITY["EPSG","4326"]],PROJECTION["Albers_Conic_Equal_Area"],PARAMETER["latitude_of_center",23],PARAMETER["longitude_of_center",-96],PARAMETER["standard_parallel_1",29.5],PARAMETER["standard_parallel_2",45.5],PARAMETER["false_easting",0],PARAMETER["false_northing",0],UNIT["metre",1,AUTHORITY["EPSG","9001"]],AXIS["Easting",EAST],AXIS["Northing",NORTH]]')
>>> rds.transform
Affine(300.0, 0.0, 1634415.0,
       0.0, -300.0, 2714805.0)
```

alexamici: > Which CF tools does this not work with?

I can't reproduce an error ATM, but CDO is usually quite picky about the content of the `coordinates` attributes.

snowman2: A big issue for not including the CRS in the coordinates is the lack of support for the `DataArray`. The CRS information will be lost for any `data_var` selected from a `Dataset` and cannot be set when creating a `DataArray` from scratch. Without the CRS information, several features in the `rio` extension won't work properly that require the CRS on the `DataArray` such as `rio.to_raster`, `rio.clip`, and `rio.reproject` to name a few.

alexamici: This is morphing into a discussion on the *xarray* data model and on the limitations of `xr.DataArray`. At them moment coordinate variables are special cased and associated with the `DataArray` while other auxiliary variables, like `grid_mapping` and `ancillary_variables ` variables `, are not.

I understand the current need for the `spatial_ref`-variable-as-coordinate hack, but I'm concerned about the fact that the resulting in-memory and on-disk data structure do not conform to the well known standard.

We should probably move the conversation on pydata/xarray.

alexamici: OK, I had completely missed https://github.com/pydata/xarray/pull/2844 and I'm reading the other issues `grid_mapping` as well, then I'll reassess.

alexamici: So technically adding support for `decode_coords="all"` in the plugin and setting `.encoding` accordingly should make everybody happy.

Note that the default would be to open with `decode_coords=True` equals to `decode_coords="coordinates"` that is similar to applying `.reset_coords`, but you can have an improved `spatial_ref`-variable-as-coordinate in-memory that also saves a correct netCDF on-disk. Cool.

alexamici:
@snowman2 The current implementation is just to see where I'd be aiming following https://github.com/pydata/xarray/pull/2844

```
>>> import xarray as xr
>>> xr.open_dataset("test/test_data/input/cog.tif", decode_coords="coordinates")  # or decode_coords=True
<xarray.Dataset>
Dimensions:      (band: 1, x: 500, y: 500)
Coordinates:
  * band         (band) int64 1
  * y            (y) float64 2.715e+06 2.714e+06 ... 2.565e+06 2.565e+06
  * x            (x) float64 1.635e+06 1.635e+06 ... 1.784e+06 1.784e+06
Data variables:
    spatial_ref  int64 ...
    band_data    (band, y, x) int16 ...

>>> xr.open_dataset("test/test_data/input/cog.tif", decode_coords="all")
<xarray.Dataset>
Dimensions:      (band: 1, x: 500, y: 500)
Coordinates:
  * band         (band) int64 1
  * y            (y) float64 2.715e+06 2.714e+06 ... 2.565e+06 2.565e+06
  * x            (x) float64 1.635e+06 1.635e+06 ... 1.784e+06 1.784e+06
    spatial_ref  int64 ...
Data variables:
    band_data    (band, y, x) int16 ...
```

I didn't manage to add the `"grid_mapping": "spatial_ref"` to `band_data.encoding` yet, but once we do it the resulting netCDF will be the same in both cases.

The best course of action would be to implement `decode_coords="all"` in `open_rasterio` though.

rioxarray/xarray_plugin.py
snowman2: @alexamici, I appreciate your thoughts and ideas on the on the matter. For now, storing the Coordinate Reference System in the coordinate versus the data variables is something I would like to discuss in a separate issue and implement any changes in a separate PR for clarity in the changeset and to allow for more thought and discussion on the subject. Would you mind removing the logic from this PR and create a separate issue?

alexamici:
@snowman2 I'll close the PR for now, waiting for a resolution of #282. Thanks for your attention!

alexamici:
Reopened the PR and renamed the engine to `"rasterio"` as this maps exactly the `open_rasterio` function except it always return a `Dataset`.

I only need to add the whatnew fragment.

alexamici:
@snowman2 I added an entry in the history file, but I have no idea how to add the reference to the documentation.

This is ready for me.

rioxarray/xarray_plugin.py
```diff
@@ -0,0 +1,42 @@
...
+
+CAN_OPEN_EXTS = {"geotif", "geotiff", "j2k", "jp2", "tif", "tiff", "vrt"}
+
+
+class RasterioBackend(xr.backends.common.BackendEntrypoint):
+    def open_dataset(
+        self,
+        filename_or_obj,
+        drop_variables=None,
+        mask_and_scale=False,
```
snowman2, line 15: You could probably mask_and_scale to `True` since this is a entrypoint. It would make it more consistent with other engines using `xarray.open_dataset`.

test/integration/test_integration_xarray_plugin.py
```diff
@@ -0,0 +1,29 @@
+import os.path
+
+import pytest
+
+from test.conftest import TEST_INPUT_DATA_DIR
+
+xr = pytest.importorskip("xarray", minversion="0.17.1.dev0")
+
+
+def test_xarray_open_dataset():
```
snowman2, line 10: Please use the `tmp_path` fixture [ref](https://docs.pytest.org/en/stable/tmpdir.html) to write the test output netcdf file.

rioxarray/xarray_plugin.py
```diff
@@ -0,0 +1,42 @@
+import os.path
+
+import xarray as xr
+
+from . import _io
+
+CAN_OPEN_EXTS = {"geotif", "geotiff", "j2k", "jp2", "tif", "tiff", "vrt"}
```
snowman2, line 7: Mind adding these as well?
```
'img', 'asc', 'png', 'jpg', 'jpeg'
```

test/integration/test_integration_xarray_plugin.py
alexamici: w00t! I didn't intend to leave the saving of the files in the code. I'll remove the calls `to_netcdf`.

rioxarray/xarray_plugin.py
alexamici: Done. And what about `masked`?

rioxarray/xarray_plugin.py
alexamici: Done

rioxarray/xarray_plugin.py
snowman2: `masked` was added before `mask_and_scale` and is there for backwards compatibility. It is mostly useful for loading in datasets that don't require scaling but masking is desired as it has simpler logic for loading in data. Being `False` by default allows the user to set `mask_and_scale=False` without also having to set `masked=False`.

docs/history.rst
```diff
@@ -10,6 +10,7 @@ Latest
 0.3.2
 -----
 - BUG: handle lock=True in open_rasterio (issue #273)
+- ENH: enable `engine="rasterio"` via xarray backend API (issue #197 pull #281)
```
snowman2, line 13: If you rebase with the changes in master from #204, you can add this to the `Lastest` section.

rioxarray/xarray_plugin.py
```diff
@@ -0,0 +1,55 @@
...
+        self,
+        filename_or_obj,
+        drop_variables=None,
+        mask_and_scale=True,
+        parse_coordinates=None,
+        lock=None,
+        masked=False,
+        variable=None,
+        group=None,
+        default_name="band_data",
```
snowman2, line 34: I saw you had `__band__` in `xarray-gdal`. Do you prefer that to `band_data`?

test/integration/test_integration_xarray_plugin.py
```diff
@@ -0,0 +1,27 @@
...
+    cog_file = os.path.join(TEST_INPUT_DATA_DIR, "cog.tif")
+
+    ds = xr.open_dataset(cog_file, engine="rasterio")
+
+    assert isinstance(ds, xr.Dataset)
+    assert "band_data" in ds.data_vars
+    assert ds.data_vars["band_data"].shape == (1, 500, 500)
+    assert "spatial_ref" not in ds.data_vars
+    assert "spatial_ref" in ds.coords
+    # uncomment after "grid_mapping" is moved into `encodings`
```
snowman2, line 20: You should be able to update this now.

rioxarray/xarray_plugin.py
alexamici: Yes, this is user-visible in *rioxarray* while it was not in my toy plugin as I was also implementing the rewrite of the band names. I vote for `no underscores` for the users.

snowman2:
> I have no idea how to add the reference to the documentation.

I think adding short code snippit referencing the new feature in `docs/getting_started/getting_started.rst` would be helpful for users to be aware of it.

alexamici:
I'm really bad a rebasing... :(

snowman2:
> I'm really bad a rebasing... :(

Hopefully this helps:

If you have this repo as the `upstream` remote:
```
git remote add upstream git@github.com:corteva/rioxarray.git
```
Then you can do:
```
git checkout master
git pull upstream master
git checkout xarray-plugin
git rebase master
```
Then follow the prompts to resolve conflicts in the files and to `git rebase --continue` until it is done.

Once completed:
```
git push -f origin xarray-plugin
```

rioxarray/xarray_plugin.py
```diff
@@ -0,0 +1,55 @@
...
+    "jpg",
+    "jpeg",
+    "png",
+    "tif",
+    "tiff",
+    "vrt",
+}
+
+
+class RasterioBackend(xr.backends.common.BackendEntrypoint):
```
snowman2, line 23: Mind adding this for future reference:
```python
    """
    .. versionadded:: 0.4
    """
```

justingruca (approved):

rioxarray/xarray_plugin.py
```diff
@@ -0,0 +1,59 @@
...
+        self,
+        filename_or_obj,
+        drop_variables=None,
+        mask_and_scale=True,
+        parse_coordinates=None,
+        lock=None,
+        masked=False,
+        variable=None,
+        group=None,
+        default_name="band_data",
```
snowman2, line 38: I noticed `**open_kwargs` is missing here. This is nice to have to be able to pass in custom arguments to `rasterio.open`.

There are two paths forward for this:
1. Only add `**open_kwargs`.
2. Remove all arguments except `filename_or_obj`, `mask_and_scale=True` and `defaut_name` arguments and use `**open_kwargs` as a catch all to pass in the rest of the arguments to `open_rasterio`.

I would be fine with either way. I am leaning towards 2 at the moment because it makes it so you don't have to update the `RasterioBackend` every time you update `open_rasterio`.

test/integration/test_integration_xarray_plugin.py
```diff
@@ -0,0 +1,25 @@
+import os.path
+
+import pytest
+
+from test.conftest import TEST_INPUT_DATA_DIR
+
+xr = pytest.importorskip("xarray", minversion="0.17.1.dev0")
```
snowman2, line 7: Does this work with `0.17.0`?

docs/getting_started/getting_started.rst
```diff
@@ -26,6 +26,12 @@ Why use :func:`rioxarray.open_rasterio` instead of `xarray.open_rasterio`?
 5. It adds the coordinate axis CF metadata.
 6. It loads raster metadata into the attributes.
 
+Starting with xarray 0.18.0 rioxarray enables passing `engine="rasterio"` to all xarray open functions:
```
snowman2, line 29:
```suggestion
rioxarray 0.4+ enables passing `engine="rasterio"` to ``xarray.open_dataset``:
```

snowman2: >  xarray 0.18.0

Shouldn't this work with 0.17+ or did I miss something?

snowman2: > rioxarray 0.4+ 

I modified it to this since it will have the pin `xarray>=0.17`

test/integration/test_integration_xarray_plugin.py
alexamici: No, it will work with the next release of *xarray*. At the moment the test runs only on `master`.

docs/getting_started/getting_started.rst
alexamici: The new backend API is now in `master` and will be in the next release that should 0.18.0.

So you need rioxarray 0.4+ and xarray 0.18+.

rioxarray/xarray_plugin.py
alexamici: The backend API insists in having all keyword arguments explicit and no `**kwargs` in order to raise user friendly error messages when an argument is not supported, so option 2 is off the table and option 1 is a bit ugly.

I'll implement option 1. ugly.

test/integration/test_integration_xarray_plugin.py
snowman2: Okay. Mind adding a comment to update `minversion="0.18"` after the next release of `xarray`?

rioxarray/xarray_plugin.py
```diff
@@ -36,6 +36,7 @@ def open_dataset(
         variable=None,
         group=None,
         default_name="band_data",
+        open_kwargs={},
```
snowman2, line 39: Since a dictionary is mutable, it is best to instead use None as the default argument and then do something like this later: `open_kwargs = {} if open_kwargs is None else open_kwargs` [ref](https://docs.python-guide.org/writing/gotchas/)

test/integration/test_integration_xarray_plugin.py
snowman2: Looks like it should work with 0.17: https://github.com/pydata/xarray/blob/v0.17.0/xarray/backends/common.py#L346

alexamici: Bits and pieces of the implementation where already present in 0.17.0, but the switch to the new plugin API got in after the release pydata/xarray#4989 (I know as I did the PR).

rioxarray/xarray_plugin.py
alexamici: This is just a passthrough will low to no change to ever be edited. I use `{}` in those trivial cases for readability, but I can change it.

test/integration/test_integration_xarray_plugin.py
snowman2: That is good to know, thanks for the clarification :+1:

snowman2:
@alexamici am I good to squash and merge? (rasterio 1.2.2 was just released and looks like it is breaking things, so I am not worried about test failures related to this PR).

alexamici:
Release should be "soonish", but not target date to my knowledge.

Anyway the changes are totally backward compatible, since che `xarray_plugin.py` is not imported in older version of xarray and will only be imported by upcoming releases.

alexamici:
You can merge now for me.

snowman2:
Thanks @alexamici :+1:

PR #283, opened 2021-04-06
Fix upstream build by pointing to the `main` branch for dask

alexamici:
Trivial fix for the rename to `main` in dask repos


snowman2:
Thanks :+1:

PR #306, opened 2021-04-20
BUG: Preserve original data type for writing to disk

### LINKED ISSUE #305: Keep in memory original data type for writing
It could be very nice to keep in memory the original data type (ie. `uint16` for landsat data or `uint8` for classified data).

Indeed, for now (unless I missed something), if we open a dataset with `masked=True` and the data is casted to float, we lose the information of the original datatype.
So we always write the dataset on disk in float, even when it is not justified (when the nodata is replaced by its true value)

snowman2:
- [x] Closes #305
 - [x] Tests added


snowman2:
This wasn't capturing changes, so opened #307.

PR #323, opened 2021-05-03
TST: Update tests for rasterio 1.2.3

snowman2:
https://github.com/mapbox/rasterio/blob/06a292dc5a30e483792d7083dc75da9c46ade50b/CHANGES.txt#L4-L16
```
Fix a regression introduced in 1.2.2: from_bounds() can again return Windows of zero height or width.
```


snowman2:
Note: Not worried about failing tests in latest - I think the changes haven't been merged into `rasterio` master branch from the 1.2 branch.

justingruca (approved):

PR #346, opened 2021-05-24
BUG: Close file handle with lock=False

### LINKED ISSUE #344: overview not read when specifying chunks or lock
#### Code Sample, a copy-pastable example if possible

```python
import rasterio
import rioxarray

# from https://openaerialmap.org/
cog_url = (
    "https://oin-hotosm.s3.amazonaws.com/"
    "5d7dad0becaf880008a9bc88/0/5d7dad0becaf880008a9bc89.tif"
)
```

```python
ds = rioxarray.open_rasterio(cog_url, lock=False, chunks=(4, "auto", -1))
ds.shape
```

    (3, 9984, 22016)

```python
dsLockChunksOverview = rioxarray.open_rasterio(cog_url, lock=False, chunks=(4, "auto", -1), overview_level=1)
dsLockChunksOverview.shape
```

    (3, 9984, 22016)

```python
dsChunksOverview = rioxarray.open_rasterio(cog_url, chunks=(4, "auto", -1), overview_level=1)
dsChunksOverview.shape
```

    (3, 9984, 22016)

```python
dsLockOverview = rioxarray.open_rasterio(cog_url, lock=False, overview_level=1)
dsLockOverview.shape
```

    (3, 9984, 22016)

```python
dsOverview = rioxarray.open_rasterio(cog_url, overview_level=1)
dsOverview.shape
```

    (3, 2496, 5504)

#### Problem description

Large COGs consequently have large overviews. It may be beneficial to read the overview in parallel without locks, as documented in [Reading COGs in Parallel](https://corteva.github.io/rioxarray/stable/examples/read-locks.html). Perhaps this behavior is documented, but I could not find it.

#### Expected Output

The correct shape of the overview `(3, 2496, 5504)` when specifying `overview_level` AND `chunks` or `lock`.

#### Environment Information

rioxarray (0.4.0) deps:
  rasterio: 1.2.3
    xarray: 0.18.2
      GDAL: 3.2.2

Other python deps:
     scipy: 1.6.3
    pyproj: 3.0.1

System:
    python: 3.7.3 (default, Jan 22 2021, 20:04:44)  [GCC 8.3.0]
executable: /usr/bin/python3
   machine: Linux-5.4.104+-x86_64-with-debian-10.9

#### Installation method
pypi

snowman2:
- [x] Tests added

For the caching file manager, it closes the file handle from the cache [ref](https://github.com/pydata/xarray/blob/071da2a900702d65c47d265192bc7e424bb57932/xarray/backends/file_manager.py#L214-L222). Since there is no cache to close with, I am thinking that the class should keep track of it like the DummyFileManager [ref](https://github.com/pydata/xarray/blob/071da2a900702d65c47d265192bc7e424bb57932/xarray/backends/file_manager.py#L316). (ping @TomAugspurger in case you have thoughts on this).

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,6 +1,7 @@
 [flake8]
 max_line_length = 88
 ignore =
+    E501 # line too long - let black handle that
     C408 # Unnecessary dict/list/tuple call - rewrite as a literal
     E203 # whitespace before ':' - doesn't work well with black
     E225 # missing whitespace around operator - let black worry about that
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -4,6 +4,7 @@ History
 Latest
 ------
 - BUG: pass kwargs with lock=False (issue #344)
+- BUG: Close file handle with lock=False (pull #346)
 
 0.4.0
 ------
--- a/rioxarray/_io.py
+++ b/rioxarray/_io.py
@@ -9,6 +9,7 @@ Source file: https://github.com/pydata/xarray/blob/1d7bcbdc75b6d556c04e2c7d7a042
 import contextlib
 import os
 import re
+import threading
 import warnings
 from distutils.version import LooseVersion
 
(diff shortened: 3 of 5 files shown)


justingruca (approved):

TomAugspurger:
I haven't looked closely at the issue / PR, but just want to note that you may want to be careful to only close the file handle when rioxarray is the one that created it. If you're passed an open file handle then typically the user should be the one to close it. Though like I said, I haven't looked closely so I don't know if that applies here.

snowman2:
Thanks for your comment @TomAugspurger :+1: 

> you may want to be careful to only close the file handle when rioxarray is the one that created it. If you're passed an open file handle then typically the user should be the one to close it.

This only changes the `URIManager` used in `rioxarray.open_rasterio`. If I understand what you are saying, I think this should be safe. Additionally, this PR makes the `close` method functional as previously it just has `pass` in there.

Issue #339, opened 2021-05-19
Using gcps with rio.reproject


oarcher:
Hi, this is my first question here:

I'd like to be able to use gcps for reproject, like rasterio is able to do.

My goal is to plot a sentinel-1 [image](https://cyclobs.ifremer.fr/static/sarwing_datarmor/xsardata/S1B_IW_GRDH_1SDV_20181013T062322_20181013T062347_013130_018428_Z010.SAFE/measurement/s1b-iw-grd-vv-20181013t062322-20181013t062347-013130-018428-001.tiff) on a map, with geoviews.

My first guess was to use:

```python
rasterize(gv.load_tiff(filename))
```

But this doesn't works, because the tiff has no transform. However, it has some gcps, and it is correctly mapped in qgis.

I was not able to find any solution to do the reprojection with rioxarray, so I've ended in a pure rasterio solution, converting to rioxarray at the end:

```python
# 100% rasterio
src = rasterio.open(filename)
gcps, src_crs = src.get_gcps()
destination = np.zeros((5000,5000)) 
dest, transform = rasterio.warp.reproject(src.read(1), destination, gcps=gcps, src_crs=src_crs, dst_crs=src_crs)

# to rioxarray
dn = xr.DataArray(dest, dims=['y','x'])
dn = dn.rio.write_crs(src_crs)
dn = dn.rio.write_transform(transform)
dn = dn.assign_coords(x=(transform * (dn.x, dn.y[0]))[0] , y=(transform * (dn.x[0], dn.y))[1]) 

# plot with geoviews
rasterize(gv.Image(dn))
```

My input image is pretty big (25000 * 15000), so ideally, the reprojection should be done with dask (#119)

snowman2:
gcps support could definitely be improved.

Using the [transform.from_gcps[(https://rasterio.readthedocs.io/en/latest/api/rasterio.transform.html#rasterio.transform.from_gcps) method in [open_rasterio](https://github.com/corteva/rioxarray/blob/5a410f5d9654430e3952d0c35d0847a01f851691/rioxarray/_io.py#L617) might be a path forward on that.

snowman2:
@mraspaud also had interest in using gcps with rioxarray.

mraspaud:
Yes I do, also for sentinel 1 data

mraspaud:
[...]

PR #370, opened 2021-07-06
ENH: kwargs & gcps support for rio.reproject

### LINKED ISSUE #369: ENH: Allow passing in arguments into `rasterio.warp.transform` in `rio.reproject`
https://rasterio.readthedocs.io/en/latest/api/rasterio.warp.html#rasterio.warp.reproject

There are additional arguments that could be passed in such as:
```
gcps (sequence of GroundControlPoint, optional) – Ground control points for the source. An error will be raised if this parameter is defined together with src_transform or rpcs.

rpcs (RPC or dict, optional) – Rational polynomial coefficients for the source. An error will be raised if this parameter is defined together with src_transform or gcps.

src_alpha (int, optional) – Index of a band to use as the alpha band when warping.

dst_alpha (int, optional) – Index of a band to use as the alpha band when warping.

num_threads (int, optional) – The number of warp worker threads. Default: 1.

init_dest_nodata (bool) – Flag to specify initialization of nodata in destination; prevents overwrite of previous warps. Defaults to True.

warp_mem_limit (int, optional) – The warp operation memory limit in MB. Larger values allow the warp operation to be carried out in fewer chunks. The amount of memory required to warp a 3-band uint8 2000 row x 2000 col raster to a destination of the same size is approximately 56 MB. The default (0) means 64 MB with GDAL 2.2.

kwargs (dict, optional) – Additional arguments passed to transformation function.
```

snowman2:
- [x] Closes #369 
 - [x] Part of #339
 - [x] Tests added

This one snowballed a bit:
```
- ENH: Allow passing in kwargs to rio.reproject (issue #369)
- ENH: Allow nodata override and provide default nodata based on dtype in rio.reproject
- ENH: Add support for passing in gcps to rio.reproject (issue #339)
```

### DIFF
--- a/rioxarray/raster_array.py
+++ b/rioxarray/raster_array.py
@@ -18,6 +18,7 @@ import rasterio
 import rasterio.mask
 import rasterio.warp
 import xarray
+from rasterio.dtypes import dtype_rev
 from rasterio.enums import Resampling
 from rasterio.features import geometry_mask
 from scipy.interpolate import griddata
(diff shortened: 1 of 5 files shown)


snowman2:
Had to pin GDAL due to #371

rioxarray/raster_array.py
```diff
@@ -87,33 +106,32 @@ def _add_attrs_proj(new_data_array, src_data_array):
...
-        ]
-        if v is not None
-    }
+
+    for key, value in (
+        ("resolution", dst_resolution),
+        ("dst_height", dst_height),
+        ("dst_width", dst_width),
+    ):
+        if key is not None:
```
justingruca, line 126: Should this be `value` instead of `key`?

snowman2: Good catch, thanks!

justingruca (approved):

PR #377, opened 2021-07-07
DOC: fix xarray page link

raybellwaves:
The docs currently links to http://xarray.pydata.org/en/stable/internals.html#extending-xarray which is not found.

This updates that to http://xarray.pydata.org/en/stable/internals/extending-xarray.html


snowman2:
Thanks :+1:

PR #392, opened 2021-08-10
BUG: Skip non-geospatial arrays in datasets when clipping

MarkusZehner:
- [x] Tests added

Hello, first thanks for the great extension for xarray objects!

I added an argument to raster_dataset.clip to enable clipping only some of the variables, as i am storing metadata along with raster data, which only use the time axis. Maybe this workaround is useful for others.
Using the raster_array.clip, i was not able to drop the data outside of the mask in place, but maybe there is a better way to fix this.

Cheers!

### DIFF
--- a/rioxarray/raster_dataset.py
+++ b/rioxarray/raster_dataset.py
@@ -243,13 +243,11 @@ class RasterDataset(XRasterBase):
         drop=True,
         invert=False,
         from_disk=False,
-        clip_vars=None,
     ):
         """
         Crops a :class:`xarray.Dataset` by geojson like geometry dicts.
 
-        .. warning:: Only works if all variables in the dataset have the same
-                     coordinates, otherwise specify them in clip_vars
+        .. warning:: Clips vars that have both x and y dims. Others are appended as is.
 
         Powered by `rasterio.features.geometry_mask`.
 
@@ -292,8 +290,6 @@ class RasterDataset(XRasterBase):
             If True, it will clip from disk using rasterio.mask.mask if possible.
             This is beneficial when the size of the data is larger than memory.
             Default is False.
-        clip_vars: :obj:`list` of :obj:`str`, optional
-            if set, only those vars will be clipped, all others are just appended to the xarray.
 
         Returns
         -------
@@ -302,14 +298,10 @@ class RasterDataset(XRasterBase):
         """
         clipped_dataset = xarray.Dataset(attrs=self._obj.attrs)
         for var in self.vars:
-            if clip_vars and var not in clip_vars:
-                clipped_dataset[var] = self._obj[var]
-            else:
+            if hasattr(self._obj[var], self.x_dim) and hasattr(self._obj[var], self.y_dim):
                 clipped_dataset[var] = (
                     self._obj[var]
-                    .rio.set_spatial_dims(
-                        x_dim=self.x_dim, y_dim=self.y_dim, inplace=True
-                    )
+                    .rio.set_spatial_dims(x_dim=self.x_dim, y_dim=self.y_dim, inplace=True)
                     .rio.clip(
                         geometries,
                         crs=crs,
@@ -319,6 +311,8 @@ class RasterDataset(XRasterBase):
                         from_disk=from_disk,
                     )
                 )
+            else:
+                clipped_dataset[var] = self._obj[var]
         return clipped_dataset.rio.set_spatial_dims(
             x_dim=self.x_dim, y_dim=self.y_dim, inplace=True
         )
--- a/rioxarray/raster_dataset.py
+++ b/rioxarray/raster_dataset.py
@@ -191,7 +191,7 @@ class RasterDataset(XRasterBase):
         )
 
     def clip_box(self, minx, miny, maxx, maxy, auto_expand=False, auto_expand_limit=3):
-        """Clip the :class:`xarray.Dataset` by a bounding box.
+        """Clip the :class:`xarray.Dataset` by a bounding box in dimensions 'x'/'y'.
 
         .. warning:: Clips variables that have dimensions 'x'/'y'. Others are appended as is.
 
@@ -249,7 +249,7 @@ class RasterDataset(XRasterBase):
         from_disk=False,
     ):
         """
-        Crops a :class:`xarray.Dataset` by geojson like geometry dicts.
+        Crops a :class:`xarray.Dataset` by geojson like geometry dicts in dimensions 'x'/'y'.
 
         .. warning:: Clips variables that have dimensions 'x'/'y'. Others are appended as is.
 
--- a/rioxarray/raster_dataset.py
+++ b/rioxarray/raster_dataset.py
@@ -218,8 +218,7 @@ class RasterDataset(XRasterBase):
         """
         clipped_dataset = xarray.Dataset(attrs=self._obj.attrs)
         for var in self.vars:
-            if hasattr(self._obj[var], self.x_dim) \
-                    and hasattr(self._obj[var], self.y_dim):
+            if self.x_dim in self._obj[var].dims and self.y_dim in self._obj[var].dims:
                 clipped_dataset[var] = (
                     self._obj[var]
                     .rio.set_spatial_dims(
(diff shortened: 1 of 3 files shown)


snowman2:
I think that this is a good idea 💡. The only improvement I can think of is to automatically determine which arrays are missing the geospatial dimensions and skip those.

You should be able to do that with `self.x_dim` and `self.y_dim`

MarkusZehner:
Yes the check for the dims is even better i think! 
The check now will only clip vars with both dims, i did not try it with vars that have either x or y (but i dont have a scenario with rasters that would need only one...)
I added the same check to clip_box, too.

MarkusZehner:
For some reason the expected sum in the tests was different on my side, but i had to manually install rasterio with pip and pyproj with conda, maybe there is a difference in the versions? Sorry for the unnecessary amount of commits, i'm still learning! ;)

rioxarray/raster_dataset.py
```diff
@@ -219,18 +218,24 @@ def clip_box(self, minx, miny, maxx, maxy, auto_expand=False, auto_expand_limit=
...
-                self._obj[var]
-                .rio.set_spatial_dims(x_dim=self.x_dim, y_dim=self.y_dim, inplace=True)
-                .rio.clip_box(
-                    minx,
-                    miny,
-                    maxx,
-                    maxy,
-                    auto_expand=auto_expand,
-                    auto_expand_limit=auto_expand_limit,
+            if hasattr(self._obj[var], self.x_dim) \
```
snowman2, line 221: I would change this to:
```python
if self.x_dim in self._obj[var].dims and self.y_dim in self._obj[var].dims:
````

**Comment by snowman2** [comment]
> Sorry for the unnecessary amount of commits, i'm still learning! ;)

No worries at all. That can help ensure that everything behaves as expected across environments. So, commit away :rocket:

File: rioxarray/raster_dataset.py (hunk 1 above)
Comment (snowman2) [reply rioxarray/raster_dataset.py:221]: Alternatively, you could do something like this:
```python
try:
    # pylint: disable=pointless-statement
    self._obj[var].rio.x_dim
    self._obj[var].rio.y_dim
except DimensionError:
    continue
```

snowman2:
Side note, setting up `pre-commit` might be helpful:
```
python -m pip install pre-commit
pre-commit install
```

MarkusZehner:
Just one thing i noticed: the ignored variables now had sometimes inconsistent chunk sizes, maybe running `clipped_dataset.chunk(chunk="auto")` could be helpful?

snowman2:
Don't worry about the failing interpolate test - a new scipy release seems to have changed the algorithm.

snowman2:
> the ignored variables now had sometimes inconsistent chunk sizes, maybe running clipped_dataset.chunk(chunk="auto") before returning could be helpful?

Unfortunately, the clip/reproject methods aren't designed/tested around dask at the moment. So, that will need to be addressed at a later time when dask support improves.

snowman2:
@MarkusZehner mind adding a test that clips a dataset with some geospatial and some non-geospatial arrays?

MarkusZehner:
> @MarkusZehner mind adding a test that clips a dataset with some geospatial and some non-geospatial arrays?

i have a mwe, using the geometries from the tests, but i am not shure how the test should look like.

```python
geometries = [
            {
                "type": "Polygon",
                "coordinates": [
                    [
                        [425499.18381405267, 4615331.540546387],
                        [425499.18381405267, 4615478.540546387],
                        [425526.18381405267, 4615478.540546387],
                        [425526.18381405267, 4615331.540546387],
                        [425499.18381405267, 4615331.540546387],
                    ]
                ],
            }
        ]

times = pd.date_range(start='2000-01-01',freq='1H',periods=6)
x = np.linspace(425493.18381405, 425532.18381405, num=6)
y = np.linspace(4615295.54054639, 4615514.54054639, num=6)
ds = xr.Dataset({
    'stuff': xr.DataArray(
                data   = np.zeros((6,6,6), dtype=float),  
                dims   = ['time', 'x', 'y'],
                coords = {'time': times,
                'x': x,
                'y':y,
                }
            )
        }
    )

ds.rio.set_spatial_dims(x_dim="x", y_dim="y", inplace=True)
ds.rio.write_crs('EPSG:26915', inplace = True)

ds['meta'] = ('time', np.random.random(6))

clipped = ds.rio.clip(geometries)

assert ds.stuff.shape == (6,6,6)
assert clipped.stuff.shape == (6,4,4)
```

snowman2:
@MarkusZehner that looks good to me. I would add:
```python
assert "meta" in clipped.data_vars
```

snowman2:
Can put the test in this file: `test/integration/test_integration_rioxarray.py`

test/integration/test_integration_rioxarray.py
```diff
@@ -298,6 +298,37 @@ def test_clip_box(modis_clip):
         assert_almost_equal(clipped_ds.rio.transform(), xdc.rio.transform())
         # make sure it safely writes to netcdf
         clipped_ds.to_netcdf(modis_clip["output"])
+        # check for variables without spatial dims
```
snowman2, line 301: Mind making this a separate test named:
`def test_clip_box__non_geospatial()`

test/integration/test_integration_rioxarray.py
```diff
@@ -298,6 +298,37 @@ def test_clip_box(modis_clip):
...
+        times = numpy.arange(6)
+        x = numpy.linspace(425493.18381405, 425532.18381405, num=6)
+        y = numpy.linspace(4615295.54054639, 4615514.54054639, num=6)
+        print(x, y)
```
snowman2, line 305: Don't forget to remove the `print`

test/integration/test_integration_rioxarray.py
```diff
@@ -484,6 +515,32 @@ def test_clip_geojson(open_func, from_disk):
             clipped_ds.rio.transform(),
             (3.0, 0.0, 425500.68381405267, 0.0, -3.0, 4615477.040546387, 0.0, 0.0, 1.0),
         )
+        # check for variables without spatial dims
```
snowman2, line 518: This looks like a duplicate of the above test. If that is the case, it can be removed. If not, mind making it a separate test?

MarkusZehner: It is the same just with `rio.clip_box` and `rio.clip`.
How about combining both into `def test_clip_clip_box__non_geospatial()`?

snowman2: I would actually prefer two separate tests for the future as it narrows down the debugging.

To reduce duplication, you could create a function that generates the dataset that you could share with the two tests.

snowman2: However, what you have is fine if you would rather not change it.

snowman2 (approved):

snowman2:
This looks great, thanks @MarkusZehner :+1:

test/integration/test_integration_rioxarray.py
MarkusZehner: I've just separated the functions, so if thats okay with you, its fine with me! 
Thanks!

snowman2:
Any issues if I squash and merge?

test/integration/test_integration_rioxarray.py
```diff
@@ -540,6 +540,64 @@ def test_clip_geojson__no_drop(open_func, invert, from_disk, expected_sum):
         assert clipped_ds.test_data.rio.nodata == xdi.rio.nodata
 
 
+def dummy_dataset_non_geospatial():
```
justingruca, line 543: It might make sense to convert this to a pytest fixture.

MarkusZehner:
Just realized different coords in geometries, fixing that ;)

test/integration/test_integration_rioxarray.py
MarkusZehner: I'm using also the same geometries as in `test_clip_geojson`, so this could also be condensed by a fixture?

MarkusZehner: How would i go about to make the constant visible for the functions without a fixture? Or just leave both in their functions?

snowman2: Justin is talking about the geometries only in this scenario. So, you would change `def geometries ...` into a global variable at the top of the file like `_CLIP_GEOMETRIES=[...]`

MarkusZehner: Ok, i see. So should this be changed to a global variable, or reverted to the geometries in the functions?

snowman2: Either way works for me :+1:

justingruca (approved):

MarkusZehner:
> Any issues if I squash and merge?

Nope :)

snowman2:
Thanks again @MarkusZehner :+1:

PR #397, opened 2021-09-13
TST: update test_estimate_utm_crs for PROJ 8.1

snowman2:
https://github.com/corteva/rioxarray/runs/3577317000?check_suite_focus=true#step:7:106

Note: changed datum from NAD83 to WGS72 because the updated version doesn't have an EPSG code for NAD83.


snowman2:
Note: Going to handle the other two failing tests separately.

justingruca (approved):

Issue #191, opened 2020-11-24
Add ability to describe color interpretation


AbelThar:
[Ability to handle color interpretation as enabled by Rasterio](https://rasterio.readthedocs.io/en/latest/topics/color.html)

Primarily ability to read existing color [with `Dataset.colorinterp`, return the sequence of `ColorInterp.<enum>`](https://rasterio.readthedocs.io/en/latest/api/rasterio.io.html?highlight=colorinterp#rasterio.io.DatasetReader.colorinterp), and the modify a Dataset by overriding it, similar to how descriptions work.

Put example of opening a dataset in `r+` just to change description of bands and their respective color implementations:
```py
with rasterio.open(dataset_path, 'r+') as src:
    src.descriptions = band_descriptions
    src.colorinterp = [rasterio.enums.ColorInterp.gray] * len(band_descriptions)
```

Probably a minor priority issue given that if you have `rioxarray` you have `rasterio` as well, but combining them both might be more readable/polished.

To be honest, have not looked as the source code to see if it is easy to implement, but would consider doing it if relatively minor.

For reference, I assume it would be similar for how descriptions work, by modifying the _xarray_ `attrs['long_name']`.

Not sure if an alternative approach exists (recently started using rioxarray).

Could not find a related issue in [either the issues page](https://github.com/corteva/rioxarray/issues?q=color) nor on [gis stackexchange](https://gis.stackexchange.com/search?q=%5Brioxarray%5D+color)

snowman2:
When reading the raster, the `photometric` option might be available in the profile and you could add that to the attributes.
Then, you could add a method to convert the `rasterio.enums.ColorInterp` to a photometric option and write that attribute to the xarray object's attributes and update `to_raster` to look for that.

Without modifying rioxarray, you can use the `photometric` option when writing to raster.

````python
rds,rio.to_raster(...,photometric="RGB")
```
Otherwise, you will have to use `rasterio` to update the `colorinterp` as `rioxarray` doesn't support the update mode.

snowman2:
Another option would be to add a `colorinterp` kwarg to `to_raster`:

```python
rds.rio.to_raster(...,colorinterp=[rasterio.enums.ColorInterp.gray] * len(band_descriptions))
```

weiji14:
Hi there :wave:, I'm interested in adding the color interpretation feature too. We're thinking of integrating `rioxarray` in PyGMT (see thread at #1555) and would be keen to help out :smile:

[...]

PR #414, opened 2021-10-02
Read colortable and colorinterp from GeoTIFF as an attribute

weiji14:
Preliminary support to read in color interpretation information from GeoTIFF files. Partially based on @alando46's code at https://github.com/pydata/xarray/pull/3136. Rasterio reference at https://rasterio.readthedocs.io/en/latest/topics/color.html

 - [ ] Addresses #191
 - [ ] Tests added

### DIFF
--- a/rioxarray/_io.py
+++ b/rioxarray/_io.py
@@ -534,7 +534,10 @@ def _get_rasterio_attrs(riods):
             attrs["units"] = riods.units
     if hasattr(riods, "colormap"):
         # A dict containing the colormap for a band
-        attrs["colormap"] = riods.colormap(1)
+        try:
+            attrs["colormap"] = riods.colormap(1)
+        except ValueError:  # NULL color table
+            pass
     if hasattr(riods, "colorinterp") and any(riods.colorinterp):
         # A tuple of the band's colorinterp property
         attrs["colorinterp"] = riods.colorinterp
--- a/rioxarray/_io.py
+++ b/rioxarray/_io.py
@@ -532,6 +532,12 @@ def _get_rasterio_attrs(riods):
             attrs["units"] = riods.units[0]
         else:
             attrs["units"] = riods.units
+    if hasattr(riods, "colormap"):
+        # A dict containing the colormap for a band
+        attrs["colormap"] = riods.colormap(1)
+    if hasattr(riods, "colorinterp") and any(riods.colorinterp):
+        # A tuple of the band's colorinterp property
+        attrs["colorinterp"] = riods.colorinterp
 
     return attrs
 
--- a/test/integration/test_integration__io.py
+++ b/test/integration/test_integration__io.py
@@ -245,6 +245,7 @@ def test_open_group_load_attrs(open_rasterio):
         assert sorted(attrs) == [
             "_FillValue",
             "add_offset",
+            "colorinterp",
             "long_name",
             "scale_factor",
             "units",
@@ -285,6 +286,7 @@ def test_open_rasterio_mask_chunk_clip():
         )
         assert attrs == {
             "add_offset": 0.0,
+            "colorinterp": (rasterio.enums.ColorInterp.gray,),
             "scale_factor": 1.0,
         }
 
--- a/test/integration/test_integration__io.py
+++ b/test/integration/test_integration__io.py
@@ -1028,6 +1028,20 @@ def test_notgeoreferenced_warning(open_rasterio):
             open_rasterio(tmp_file)
 
 
+def test_geotiff_attr_loading(open_rasterio):
+    with open_rasterio(
+        "https://oceania.generic-mapping-tools.org/server/earth/earth_day/earth_day_01d_p.tif"
+    ) as rds:
+        rds = rds.band_data if hasattr(rds, "band_data") else rds
+
+        assert rds.dims == ("band", "y", "x")
+        assert rds.shape == (1, 180, 360)
+        # assert rds.attrs["scale_factor"] == 1.0
+        # assert rds.attrs["add_offset"] == 0.0
+        assert isinstance(rds.attrs["colormap"], dict)
+        assert rds.attrs["colorinterp"] == (rasterio.enums.ColorInterp.palette,)
+
+
 @pytest.mark.xfail(
     version.parse(rasterio.__gdal_version__) < version.parse("3.0.4"),
     reason="This was fixed in GDAL 3.0.4",


test/integration/test_integration__io.py
```diff
@@ -285,6 +286,7 @@ def test_open_rasterio_mask_chunk_clip():
         )
         assert attrs == {
             "add_offset": 0.0,
+            "colorinterp": (rasterio.enums.ColorInterp.gray,),
```
snowman2, line 289: This will need to be converted to either a string representation or the integer representation. The reason for this is to enable calling `to_netcdf` without serialization issues.

Here is an example converting to the integer enumeration value:
```python
>>> import rasterio
>>> rasterio.enums.ColorInterp.gray
<ColorInterp.gray: 1>
>>> rasterio.enums.ColorInterp.gray.value
1
```

snowman2: An alternative would be to read/write the `photometric` creation option of the profile into/from the attributes. This is always in the string format and shouldn't have any serialization issues.

rioxarray/_io.py
```diff
@@ -532,6 +532,15 @@ def _get_rasterio_attrs(riods):
             attrs["units"] = riods.units[0]
         else:
             attrs["units"] = riods.units
+    if hasattr(riods, "colormap"):
+        # A dict containing the colormap for a band
+        try:
+            attrs["colormap"] = riods.colormap(1)
```
snowman2, line 538: This only gets the colormap for the first band. I imagine it would be good to handle this for all of the bands. [docs](https://rasterio.readthedocs.io/en/latest/topics/color.html#reading-colormaps)

test/integration/test_integration__io.py
weiji14: Actually, the `colorinterp` enum can be serialized since it has an integer representation. It's the `colormap` dict which I'm having trouble with. The `colormap` dict looks something like this:

```python
{0: (248, 224, 168, 255),
 1: (240, 224, 152, 255),
 2: (104, 104, 96, 255),
 ...
 255: (248, 248, 248, 255)}
```
Do you think we should keep this `colormap` as an xarray attr, or should it be loaded in the data_var as 4 bands (i.e. RGBA)?

rioxarray/_io.py
weiji14: Yep, probably need to loop through all the bands, but need to think about https://github.com/corteva/rioxarray/pull/414/files#r722041137 first :slightly_smiling_face:

test/integration/test_integration__io.py
snowman2: Since it is a dict, you could use `json.dumps` & `json.loads` to convert to/from a string.

weiji14: Ok, the json serialization has been implemented in 54f9560a883fcfbbbc83a67b5abf7d431cca4384. Had to do some extra magic because the dict keys have to be `int`, but JSON only allows `str` keys. I did try an alternative method using yaml in 5b07d7fdc7b5a7d38fef89bca5795f26077ca710 as suggested by https://stackoverflow.com/questions/17099556/why-do-int-keys-of-a-python-dict-turn-into-strings-when-using-json-dumps#comment24737635_17099556, but maybe will stick with JSON.

weiji14: > Do you think we should keep this `colormap` as an xarray attr, or should it be loaded into the data_var as 4 bands (i.e. RGBA)?

Just to follow up on the data_var loading, I tried converting the [`earth_day_01d_p.tif`](https://oceania.generic-mapping-tools.org/server/earth/earth_day/earth_day_01d_p.tif) GeoTiff to NetCDF directly using `gdal_translate -of NetCDF -expand rgba earth_day_01d_p.tif earth_day_01d_p.nc`. 

This is the output of `gdalinfo earth_day_01d_p.nc`

```
Driver: netCDF/Network Common Data Format
Files: earth_day_01d_p.nc
Size is 512, 512
Metadata:
  NC_GLOBAL#Conventions=CF-1.5
  NC_GLOBAL#GDAL=GDAL 3.3.2, released 2021/09/01
  NC_GLOBAL#GDAL_AREA_OR_POINT=Area
  NC_GLOBAL#history=Sat Nov 06 23:41:47 2021: GDAL CreateCopy( earth_day_01d_p.nc, ... )
Subdatasets:
  SUBDATASET_1_NAME=NETCDF:"earth_day_01d_p.nc":Band1
  SUBDATASET_1_DESC=[180x360] Band1 (8-bit integer)
  SUBDATASET_2_NAME=NETCDF:"earth_day_01d_p.nc":Band2
  SUBDATASET_2_DESC=[180x360] Band2 (8-bit integer)
  SUBDATASET_3_NAME=NETCDF:"earth_day_01d_p.nc":Band3
  SUBDATASET_3_DESC=[180x360] Band3 (8-bit integer)
  SUBDATASET_4_NAME=NETCDF:"earth_day_01d_p.nc":Band4
  SUBDATASET_4_DESC=[180x360] Band4 (8-bit integer)
Corner Coordinates:
Upper Left  (    0.0,    0.0)
Lower Left  (    0.0,  512.0)
Upper Right (  512.0,    0.0)
Lower Right (  512.0,  512.0)
Center      (  256.0,  256.0)
```
So the bands have been expanded into NetCDF subdatasets. I then loaded `earth_day_01d_p.nc` using rioxarray.open_rasterio, and it produces the following:

```
<xarray.Dataset>
Dimensions:  (y: 180, x: 360, band: 1)
Coordinates:
  * y        (y) float64 89.5 88.5 87.5 86.5 85.5 ... -86.5 -87.5 -88.5 -89.5
  * x        (x) float64 -179.5 -178.5 -177.5 -176.5 ... 176.5 177.5 178.5 179.5
  * band     (band) int64 1
    crs      int64 0
Data variables:
    Band1    (band, y, x) uint8 ...
    Band2    (band, y, x) uint8 ...
    Band3    (band, y, x) uint8 ...
    Band4    (band, y, x) uint8 ...
Attributes:
    Conventions:         CF-1.5
    GDAL:                GDAL 3.3.2, released 2021/09/01
    GDAL_AREA_OR_POINT:  Area
    history:             Sat Nov 06 23:41:47 2021: GDAL CreateCopy( earth_day...
```

Notice the Band1-Band4 data variables. So I'm thinking there should be a way to mimic this structure. Need to find a way to use the colormap dict currently in the attr, and get these Bands 1/2/3/4 information out as data variables.

snowman2:
#429 should help.

rioxarray/raster_writer.py
```diff
@@ -56,6 +57,16 @@ def _write_metatata_to_raster(raster_handle, xarray_dataset, tags):
         if add_offset is not None:
             raster_handle.offsets = (add_offset,) * raster_handle.count
 
+    # write colorinterp and colormap
+    # see https://rasterio.readthedocs.io/en/latest/topics/color.html
+    try:
+        raster_handle.colorinterp = tags["colorinterp"]
```
snowman2, line 63: Mind using separate try/except blocks for colorinterp and colormap?

test/integration/test_integration__io.py
snowman2: Mind doing `ncdump -h earth_day_01d_p.nc`  and pasting the output here?

snowman2:
Closing because this PR has become stale. Feel free to re-open or submit a new PR if you would like to contribute this feature.

PR #490, opened 2022-03-25
Remove crs after rio crs

### LINKED ISSUE #488: `xarray.Dataset.reproject` doesn't modify the crs attribute
#### Code Sample, a copy-pastable example if possible

```python
print("target crs: ", best_crs)
print("current crs: ", ds.crs)
reprojected = ds.rio.reproject(best_crs, resampling=0)
print("crs after reprojection: ", reprojected.crs)
```
The code above prints

```
target crs:  +init=epsg:32631
current crs:  +init=epsg:32630
crs after reprojection:  +init=epsg:32630
```
#### Problem description

I would expect that the dataset being output after the reprojection has a modified CRS. Since this is not happening the reality of the data and the metadata are out of sync causing problems downstream. This behaviour is also observed on `merge_datasets` which will generate a dataset with nominally the CRS of the first dataset that it received, regardless of the `crs` parameter.

That being said, the output looks as it has been correctly reprojected.

#### Expected Output

A reprojected dataset with a CRS modified to be that of the destiny CRS.

#### Environment Information

rioxarray (0.10.2) deps:
  rasterio: 1.2.10
    xarray: 2022.3.0
      GDAL: 3.3.2

Other python deps:
     scipy: 1.7.1
    pyproj: 3.2.1

System:
    python: 3.9.7 (default, Sep 15 2021, 08:09:16)  [GCC 10.3.0]
executable: /home/guillem/.pyenv/versions/3.9.7/envs/GISEnv/bin/python3.9
   machine: Linux-5.11.0-7633-generic-x86_64-with-glibc2.33

#### Installation method
 - pip

GCBallesteros:
- [ x] Closes #488
 - [ x] Fully documented, including `docs/history.rst` for all changes and `docs/rioxarray.rst` for new API

### DIFF
--- a/rioxarray/rioxarray.py
+++ b/rioxarray/rioxarray.py
@@ -313,6 +313,11 @@ class XRasterBase:
         crs = crs_from_user_input(input_crs)
         obj = self._get_obj(inplace=inplace)
         obj.rio._crs = crs
+
+        # Remove the crs from the underlying xarray to avoid confusion and force
+        # the user to manipulate the rioxarray controlled one.
+        if "crs" in self._obj.attrs:
+            del self._obj.attrs["crs"]
         return obj
 
     @property
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -4,6 +4,10 @@ History
 Latest
 -------
 
+0.10.3
+------
+- BUG: Remove underlying xarray `crs` once rioxarray's one is ready (issue #488)
+
 0.10.2
 -------
 - BUG: Lazy load colormap through _manager.acquire() in merge (issue #479)


snowman2:
Thanks @GCBallesteros :+1:

I see that you added it to `set_crs`. Would you mind moving it to `write_crs` ?

snowman2:
Also, mind adding a simple test around here: https://github.com/corteva/rioxarray/blob/2888deb1e2abeccf1b7b531378a58762d40685aa/test/integration/test_integration_rioxarray.py#L2021

It just needs to have a Dataframe with the `crs` attribute, call `rio.write_crs` and check that the `crs` attribute is gone?

rioxarray/rioxarray.py
```diff
@@ -447,6 +448,12 @@ def write_crs(self, input_crs=None, grid_mapping_name=None, inplace=False):
             )
         data_obj.coords[grid_mapping_name].rio.set_attrs(grid_map_attrs, inplace=True)
 
+        # remove old crs if exists
+        try:
```
snowman2, line 452: This could be simplified with:
```python
data_obj.attrs.pop("crs", None)
```

test/integration/test_integration_rioxarray.py
```diff
@@ -2024,6 +2024,13 @@ def test_get_crs_dataset():
     assert test_ds.encoding["grid_mapping"] == "spatial_ref"
     assert test_ds.rio.crs.to_epsg() == 4326
 
+def test_crs_is_removed():
+    test_ds = xarray.Dataset(attrs=dict(crs="+init=epsg:4326"))
+    test_ds = test_ds.write_crs(4326)
+
+    expected_exception = AttributeError
+    with pytest.raises(expected_exception):
```
snowman2, line 2032: You could simplify this with: `assert "crs" not in test_ds.attrs`

GCBallesteros:
A number of tests are failing now due to the removal of the crs attribute. Should I fix them by making use of the `rio.crs` attribute?

snowman2:
> A number of tests are failing now due to the removal of the crs attribute. Should I fix them by making use of the rio.crs attribute?

For `test_merge_arrays__res` and `test_merge_arrays` I would remove the `crs` attribute usage in the tests entirely as it will no longer be found.

For `test_reproject__grid_mapping` I would suggest replacing `_get_attr(mda, "crs")` with `mda.rio.crs` and `_get_attr(mdc, "crs")` with `mdc.rio.crs`.

snowman2:
I pushed a fix for the scenario when the CRS isn't found by rasterio/GDAL, but can be determined from the attributes by rioxarray.

justingruca (approved):

PR #504, opened 2022-04-12
Add URL for rasterio-docs/merge API

chbrandt:
In rioxarray.merge docs, the reference to "rasterio docs" is helpful but still quite vague; This adds the actual URL for them.

### DIFF
--- a/rioxarray/merge.py
+++ b/rioxarray/merge.py
@@ -132,7 +132,7 @@ def merge_arrays(
     precision: float, optional
         Number of decimal points of precision when computing inverse transform.
     method: str or callable, optional
-        See rasterio docs.
+        See rasterio docs (https://rasterio.readthedocs.io/en/stable/api/rasterio.merge.html).
     crs: rasterio.crs.CRS, optional
         Output CRS. If not set, the CRS of the first DataArray is used.
     parse_coordinates: bool, optional
--- a/rioxarray/merge.py
+++ b/rioxarray/merge.py
@@ -132,7 +132,7 @@ def merge_arrays(
     precision: float, optional
         Number of decimal points of precision when computing inverse transform.
     method: str or callable, optional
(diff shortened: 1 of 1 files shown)


rioxarray/merge.py
```diff
@@ -132,7 +132,7 @@ def merge_arrays(
     precision: float, optional
         Number of decimal points of precision when computing inverse transform.
     method: str or callable, optional
-        See rasterio docs.
+        See rasterio docs (https://rasterio.readthedocs.io/en/stable/api/rasterio.merge.html).
```
djhoese, line 135: I'm not familiar with what rioxarray uses for sphinx docs, but theoretically you could put

```
:mod:`rasterio.merge`
```

and when rendered to the sphinx HTML is should automatically map to the API docs in rasterio. This won't show up in the raw string docstring, but maybe something @snowman2 is interested in.

snowman2: As @djhoese mentioned, you could follow the pattern shown here: https://github.com/corteva/rioxarray/blob/94d3d53d0e3bfd12a7c3d1ce7fffd91fc7f5e141/rioxarray/raster_dataset.py#L89

And do something like this:
```
See :func:`rasterio.merge.merge` for more details.
```

snowman2:
Thanks for catching this @chbrandt :+1:

rioxarray/merge.py
snowman2: @chbrandt, what are your thoughts on modifying this PR with the suggestions above?

chbrandt: Agree, @snowman2

snowman2:
Looks good. Would you like to squash your commits or do you mind if I squash on merge?

chbrandt:
Hi @snowman2 , yes, please, a squash on merge is very welcome. Thank you.

snowman2:
Thanks @chbrandt :+1:

PR #534, opened 2022-06-16
MNT: corrections to setup.cfg; use "pytest"; add "make dist" target

mwtoews:
This PR is very similar to https://github.com/corteva/geocube/pull/111

- `setup.cfg`:
  - remove `bdist_universal` option
  - Use `license_files` to side-step warning in `python -m build`
  - Specify `long_description_content_type` to side-step warning in `twine check dist/*`
  - Add dynamic version from attribute
- `setup.py` is now a basic shim file, and does nothing custom; it might be removed at some point
- Use `pytest` command instead of `py.test` ([ref](https://stackoverflow.com/a/41893170/))
- Add `make dist` target, which was referenced in the Makefile
- Add `twine check --strict dist/*` before uploading release -- also `.whl` files could and should be uploaded to PyPI
- There is no "recusive-prune" option for MANIFEST.in -- just "prune"

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,11 +2,13 @@
 author = rioxarray Contributors
 author_email = alansnow21@gmail.com
 name = rioxarray
+version = attr: rioxarray._version.__version__
 description = geospatial xarray extension powered by rasterio
 keywords = rioxarray, xarray, rasterio
 long_description = file: README.rst
+long_description_content_type = text/x-rst
 license = Apache
-license_file = LICENSE
+license_files = LICENSE
 platform = any
 classifiers =
     Development Status :: 4 - Beta
@@ -72,6 +74,3 @@ all =
     %(doc)s
     %(test)s
     %(dev)s
-
-[bdist_wheel]
-universal = 1
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -148,7 +148,7 @@ jobs:
         shell: bash
         run: |
           source activate test
-          py.test --cov-report term-missing --cov=rioxarray --cov-report xml
+          pytest --cov-report term-missing --cov=rioxarray --cov-report xml
 
       - name: Test Build docs
         shell: bash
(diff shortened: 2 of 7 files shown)


setup.cfg
```diff
@@ -2,11 +2,13 @@
 author = rioxarray Contributors
 author_email = alansnow21@gmail.com
 name = rioxarray
+version = attr: rioxarray._version.__version__
```
snowman2, line 5: Due to importing methods to the main `rioxarray` namespace, this is not possible as it calls external dependencies potentially not installed yet. It will have to be reverted to the previous method for retrieving `__init__.py`.

mwtoews: This normally isn't an issue with `python -m build` (#536) which uses a venv isolated environment by default to gather the specified dependencies from `install_requires`. The issue you are describing can only be triggered using `python -m build --no-isolation` to show (e.g.) `ModuleNotFoundError: No module named 'numpy'`.

snowman2: I wonder if this is the line causing the troubles in the docker build: https://github.com/corteva/rioxarray/blob/39c918930ed2694e979cd557556f9a28dc296ae4/.github/workflows/tests.yaml#L80

mwtoews: hmm, perhaps I'll need to investigate this a bit further...

mwtoews: yes, I was able to reproduce this, but essentially all "python setup.py" commands are more-or-less deprecated and should be phased out (e.g. [here](https://stackoverflow.com/questions/72468946/setuptools-pyproject-toml-equivalent-to-python-setup-py-clean-all)). Was there a specific reason for it?

snowman2: Sounds like it isn't necessary, so you can remove it.

mwtoews:
I'll assume that "Test latest dependencies" is unrelated. Merge #536 before this one, and I'll merge it here too to get the build test.

snowman2:
Looks great 👍. Did you want to squash your commits or do you mind if I do it on merge?

mwtoews:
Yes, squash with title-only is good, as a xref to the RR is sufficient. Thanks!

snowman2:
Thanks @mwtoews 👍

PR #536, opened 2022-06-18
CI: add test for building sdist and wheel packages

mwtoews:
This CI workflow essentially checks `python -m build` to see if it builds sdist and wheel packages.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -208,3 +208,29 @@ jobs:
         run: |
           source activate test
           py.test --cov-report term-missing --cov=rioxarray --cov-report xml
+
+
+  test_build:
+    needs: linting
+    name: Test building
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -208,3 +208,29 @@ jobs:
...
+
+      - name: Build sdist and wheel packages
+        shell: bash
+        run: |
+          python -m build
+
+      - name: Check packages
+        shell: bash
+        run: |
+          twine check dist/*
```
snowman2, line 236: Do you think that the `--strict` flag would add any value here?

snowman2:
This will be nice to have, thanks 👍

.github/workflows/tests.yaml
mwtoews: yes, planning to enable it after #534, otherwise the current warnings will fail the workflow.

snowman2:
Thanks @mwtoews 👍

PR #541, opened 2022-06-30
TYPE: Address AreaOfInterest type complaints

snowman2:


snowman2:
Thanks @justingruca :+1:

PR #569, opened 2022-08-30
DOC: add notes to docs that xarray.open_rasterio is deprecated

JessicaS11:
For clarity, alert the user that `xarray.open_rasterio` is now deprecated and should not be used.


snowman2:
Thanks @JessicaS11 :+1:

PR #600, opened 2022-11-01
ENH: Added band_as_variable option to open_rasterio

### LINKED ISSUE #81: Store band specific attributes in coordinates
Currently there are band-specific attributes stored in the `attrs`. However, this can be potentially lossy when performing xarray operations. So, I am thinking about storing them in the coordinates as arrays along the `band` dimension. This will also be useful when selecting a specific band as you will only have the subset of data associated with the band.

Proposed coordinates:
- ~~nodatavals~~
- scales
- offsets
- descriptions (maybe)
- units (maybe)

Additionally. I am thinking about adding a `profile` and or `metadata` coordinate to store the `profile` information from the raster not stored in other attributes to persist when writing. For example - compression, tiling, etc...

This should make things simpler and more reliable for #76.

cc: @alfredoahds @justingruca

### LINKED ISSUE #296: ENH: Add ability to load bands as data_vars
@alexamici brought this up in #197 and has been something I have thought about for a while and haven't come up with a great solution yet.

`xarray-gdal` implementation: https://github.com/bopen/xarray-gdal/blob/main/xarray_gdal/xarray_plugin.py

Benefits:
- It could potentially enable round-trip loading in of data as in #249
- Would provide a nice mechanism to store band specific attributes in each variable (possible fix for #81)

Why hasn't it been done already?
- Backwards compatibility/consistency with `xarray.open_rasterio`
- Loading it as a `DataArray` allows `open_rasterio` to work with multidimensional files where the band is a coordinate (`time`).
- The coordinate mechanism provides a nice interface for selecting the band you are interested in by an index `rds.sel(band=1)`.

It may make sense to have an option such as `band_as_variable` where you can toggle this mode on.

snowman2:
- [x] Closes #296
 - [x] Closes #81
 - [x] Tests added

### DIFF
--- a/rioxarray/_io.py
+++ b/rioxarray/_io.py
@@ -7,6 +7,7 @@ Source file: https://github.com/pydata/xarray/blob/1d7bcbdc75b6d556c04e2c7d7a042
 """
 
 import contextlib
+import functools
 import importlib.metadata
 import os
 import re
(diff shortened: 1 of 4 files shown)


rioxarray/_io.py
```diff
@@ -711,6 +796,49 @@ def _load_subdatasets(
...
+    """
+    global_tags = _parse_tags(riods.tags())
+    data_vars = {}
+    for band in riods.indexes:
+        band_riods = SingleBandDatasetReader(
+            riods=riods,
+            bidx=band - 1,
+        )
+        band_name = f"band_{band}"
+        data_vars[band_name] = (
```
clausmichele, line 822: Why not leaving the possibility to use band description as band name if it is available?

snowman2: The variable name should only contain alphanumeric characters and underscores. The band description could potentially be a sentence with any characters. This ensures consistency and stability.

clausmichele: Fair. Then in my scenario I will just use the variables/band metadata and set their names accordingly.
Important is that we parse also the bands metadata and now seems fine!

clausmichele (approved):

justingruca (approved):

snowman2:
Thanks @clausmichele @justingruca :+1:

PR #606, opened 2022-11-11
BUG: Add GDAL 3.6 driver auto-select fix

snowman2:
Related: https://github.com/rasterio/rasterio/pull/2634

Issue #544, opened 2022-07-05
can only read 32 layers from .hdf files before returning a `FileNotFound` error


jamie-sgro:
<!-- Please search existing issues to avoid creating duplicates. -->


#### Code Sample, a copy-pastable example if possible

I've created a small repo with the necessary code to recreate the below error:
https://github.com/jamie-sgro/xarray-recreate-bug

#### Problem description

In Docker environments only, throws the below error. This only occurs
when trying to read .hdf files with a cumulative total of >32 layers. It always fails on the 33rd layer being read into memory regardless of the order of the files
and the contents of the files themselves. Note we use a copy of a file
for each iteration and it still fails

```
rasterio.errors.RasterioIOError: HDF4_EOS:EOS_GRID:/tmp/pytest-of-root/
pytest-5/test_can_open_hdf4_closer_to_e0/file3:MODIS_Grid_16DAY_1km_VI:1
km 16 days blue reflectance: No such file or directory
```

I believe this is an error in the intersection between xarray, rioxarray, and rasterio. See these two other issues for more details:
- https://github.com/pydata/xarray/issues/6715
- https://github.com/rasterio/rasterio/issues/2490

<details>
<summary>Full Error</summary>

```
Last login: Tue Jul  5 12:28:09 on ttys003
docker exec -it 9763aa865198baad81e9e25fd70580f20cb3d4fb0b83ef64edc2f3fba60c9e92 /bin/sh
(base) jamiesgro@Jamies-MacBook-Pro ~ % docker exec -it 9763aa865198baad81e9e25fd70580f20cb3d4fb0b83ef64edc2f3fba60c9e92 /bin/sh
# pytest
========================================================================================================================================== test session starts ==========================================================================================================================================
platform linux -- Python 3.9.2, pytest-7.1.2, pluggy-1.0.0
rootdir: /app
collected 3 items                                                                                                                                                                                                                                                                                       

tests/test_rasterio_open.py .                                                                                                                                                                                                                                                                     [ 33%]
tests/test_xarray_open_hdf4.py .F                                                                                                                                                                                                                                                                 [100%]

=============================================================================================================================================== FAILURES ================================================================================================================================================
____________________________________________________________________________________________________________________________________ test_using_xarray_via_rioxarray ____________________________________________________________________________________________________________________________________

>   ???

rasterio/_base.pyx:261: 
[...]

PR #607, opened 2022-11-11
BUG: Fix closing files manually

### LINKED ISSUE #485: How to overwrite raster opened with rioxarray.open_rasterio ?
using rioxarray.open_rasterio and subsequently rio.to_raster to edit a raster file raises a permission denied error. I can't figure out any way make changes to a TIF and then overwrite it again. Calling .close() on the dataarray seems to have no effect either.

```
tif = rioxarray.open_rasterio(tifpath)
tif.close()
tif.rio.to_raster(tifpath)
```

`CPLE_AppDefinedError: Deleting ....tif failed: Permission denied`

Expected behavior:
overwrite the file, and tif.close() statement to be unnecessary. 

As I understand it, xarray reads serialized data lazily, which is why open_rasterio would leave the file open? If that's not accurate, then open_rasterio should either close the file itself, or be implemented as a context manager. Otherwise, it makes sense that rio.to_raster() should be closing the file before overwriting it. 

As of now, the only way I've been able to edit a raster is to save as a different file, close the python kernel, and manually copy the new file over, which defeats the purpose of using python in the first place. 

I'm on windows FYI
rioxarray 0.9.0
xarray 0.20.0

snowman2:
In #606 it became evident that the files were not being closed. This fix should help users to close Datasets that have data_vars pointing to files.

 - [x] Related #485
 - [x] Related #544

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -26,8 +26,8 @@ jobs:
   docker_tests:
     needs: linting
     runs-on: ubuntu-latest
-    name: Docker | python=${{ matrix.python-version }} | rasterio${{ matrix.rasterio-version }} | scipy ${{ matrix.run-with-scipy }}
-    container: osgeo/gdal:ubuntu-full-3.4.2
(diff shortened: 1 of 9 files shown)


justingruca (approved):

snowman2:
Thanks @justingruca :+1:

PR #619, opened 2022-12-14
BUG: Handle additional data types in `rio.reproject`

### LINKED ISSUE #618: `_NODATA_DTYPE_MAP` causing bug in `rio.reproject()`
## Example code

`rio.reproject` generates errors when the input data array is `int64`.

```python
import xarray as xr
import rioxarray
from rasterio.enums import Resampling

da = xr.DataArray(
        [[1, 2, 3], [4, 5, 6]],
        dims=("y", "x"),
        coords={"y": [1.5, 0.5], "x": [0.5, 1.5, 2.5]},
    ).rio.write_crs("EPSG:4326").astype("int64")

# da.rio.set_nodata(-99, inplace=True) solves the issue

da.rio.reproject(
    da.rio.crs,
    shape = (2 * da.rio.height, 2 * da.rio.width),
    resampling = Resampling.nearest
)
```

## Cause

`rasterio.dtypes.dtype_rev` now [returns](https://github.com/rasterio/rasterio/blob/84f3bbad70dca3767b2c8e6b061aecc008b3efe3/rasterio/dtypes.py#L47) keys 12 and 13 for `int64` and `uint64` respectively, and 14 for `int8` (if GDAL is above a certain version). However, the above example generates a key error because [`_NODATA_DTYPE_MAP`](https://github.com/corteva/rioxarray/blob/0f77e13b54ba82ffb9cf2f655c9e310f85890bd9/rioxarray/raster_array.py#L70) needs to be updated to include these new types, and currently only has keys to 1 to 11. The version of the [GDAL reference table originally used](https://github.com/OSGeo/gdal/blob/cde27dc7641964a872efdc6bbcf5e3d3f7ab9cfd/gdal/swig/python/gdal-utils/osgeo_utils/gdal_calc.py#L62) has been [updated](https://github.com/OSGeo/gdal/blob/dee861e7c91c2da7ef8ff849947713e4d9bd115c/swig/python/gdal-utils/osgeo_utils/gdal_calc.py#L61) with defaults for these new types (although they are all `None`).

Note that there is another difference between the newer reference table and what's in `_NODATA_DTYPE_MAP`. The `cint` and `cfloat` default values are now `None`.

## Proposal

Easiest proposal would be to simply update `_NODATA_DTYPE_MAP` to include 12, 13, and 14, and adjust `cint` and `cfloat`. It would now be (followed `CONTRIBUTING.rst`, but don't have permission to push and make PR):

```python
_NODATA_DTYPE_MAP = {
    1: 255,  # GDT_Byte
    2: 65535,  # GDT_UInt16
    3: -32768,  # GDT_Int16
    4: 4294967293,  # GDT_UInt32
    5: -2147483647,  # GDT_Int32
    6: 3.402823466e38,  # GDT_Float32
    7: 1.7976931348623158e308,  # GDT_Float64
    8: None,  # GDT_CInt16
    9: None,  # GDT_CInt32
    10: 3.402823466e38,  # GDT_CFloat32
    11: 1.7976931348623158e308,  # GDT_CFloat64
    12: None,  # GDT_Int64
    13: None,  # GDT_UInt64
    14: None,  # GDT_Int8
}
```

However, I am unaware why this is necessary (although not an expert in these things). If the existing nodata value is `None`, why only put to default during reprojection? Maybe worth considering dropping this to prevent needing to manually align again in the future?

caldwellst:
- [ ] Closes #618
 - [ ] Documented in `docs/history.rst` and edited hyperlink comment in `rioxarray/raster_array.py`

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -3,6 +3,7 @@ History
 
 Latest
 ------
+- BUG: Handle data type error in `rio.reproject` (issue #618)
 
 0.13.2
 ------
--- a/rioxarray/raster_array.py
+++ b/rioxarray/raster_array.py
@@ -65,8 +65,8 @@ _RASTERIO_GTE_12 = version.parse(
 
 # DTYPE TO NODATA MAP
 # Based on: https://github.com/OSGeo/gdal/blob/
-# cde27dc7641964a872efdc6bbcf5e3d3f7ab9cfd/gdal/
-# swig/python/gdal-utils/osgeo_utils/gdal_calc.py#L62
+# dee861e7c91c2da7ef8ff849947713e4d9bd115c/
+# swig/python/gdal-utils/osgeo_utils/gdal_calc.py#L61
 _NODATA_DTYPE_MAP = {
     1: 255,  # GDT_Byte
     2: 65535,  # GDT_UInt16
--- a/rioxarray/raster_array.py
+++ b/rioxarray/raster_array.py
@@ -513,7 +513,7 @@ class RasterArray(XRasterBase):
 
     def _get_dst_nodata(self, nodata: Optional[float]) -> Optional[float]:
         default_nodata = (
-            _NODATA_DTYPE_MAP[dtype_rev[self._obj.dtype.name]]
+            _NODATA_DTYPE_MAP.get(dtype_rev[self._obj.dtype.name])
             if self.nodata is None
             else self.nodata
         )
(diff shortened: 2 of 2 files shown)


snowman2:
One option to make this more future proof is to change this: https://github.com/corteva/rioxarray/blob/0f77e13b54ba82ffb9cf2f655c9e310f85890bd9/rioxarray/raster_array.py#L513

to:
```python
_NODATA_DTYPE_MAP.get(dtype_rev[self._obj.dtype.name])
```
Thoughts?

caldwellst:
I think that sounds like a solid idea, particularly as it seems unlikely new data types will be added with corresponding defaults. It may be useful to just set a reminder to check in a set period of time (1 year) if that list has changed in case standard defaults have been added for those that are `None`.

snowman2:
> I think that sounds like a solid idea

Mind updating this PR with the change?

snowman2:
Thanks @caldwellst :+1:

PR #620, opened 2022-12-15
CI: Update to GDAL 3.6.1 for tests

snowman2:


snowman2:
Note: GDAL test failures due to related to https://github.com/rasterio/rasterio/issues/2662

PR #621, opened 2022-12-19
MNT: Add dockerfile for running tests with different GDAL versions

snowman2:
Found this to be useful recently with updating GDAL 3.6.1 tests.


justingruca (approved):

snowman2:
Thanks @justingruca :+1:

PR #628, opened 2023-01-16
DOC: Document grid_mapping in encoding

stefank0:
I think there should be some more documentation about writing the grid_mapping to the encoding, for example to avoid the following issue: ds.to_netcdf(path, encoding=encoding) overwrites the grid_mapping encoding and results in incomplete CRS information (results in a NetCDF for which the CRS is not automatically recognized by QGIS, for example). Probably this is commonly used to set compression (accepted answer to https://stackoverflow.com/questions/40766037/specify-encoding-compression-for-many-variables-in-xarray-dataset-when-write-to).

 - [ ] Closes #xxxx
 - [ ] Tests added


snowman2:
Related: https://github.com/pydata/xarray/discussions/5709

snowman2:
Thanks @stefank0 :+1:

PR #673, opened 2023-05-22
DOC: Add notes about pyresample & odc-geo for re-projecting with dask

snowman2:
Related #119


justingruca (approved):

snowman2:
Thanks @justingruca :+1:

PR #675, opened 2023-05-26
DOC: Improved transform bounds example to demonstrate why it is needed

snowman2:
See it rendered here: https://github.com/snowman2/rioxarray/blob/transform_bounds/docs/examples/transform_bounds.ipynb


justingruca (approved):

snowman2:
Thanks @justingruca :+1:

PR #684, opened 2023-07-15
CI: don't fetch appache-arrow repo

Kirill888:
In GDAL 3.5.3 docker repo keys for appach-arrow are stale, this will probably be fixed by rebuilding that docker image. For now we can just remove that repo before running update.

extracting build fix into a separate PR from #683 .


snowman2:
Thanks @Kirill888 👍

PR #765, opened 2024-04-02
REF: Use rasterio.io.MemoryFile in merge

snowman2:
Related: https://github.com/corteva/rioxarray/discussions/204#discussioncomment-269328

Failing tests: https://github.com/corteva/rioxarray/actions/runs/8126712662/job/22210778353

```
FAILED test/integration/test_integration_merge.py::test_merge_arrays[True] - AssertionError: 
Arrays are not almost equal to 7 decimals
 ACTUAL: <xarray.DataArray '__xarray_dataarray_variable__' ()> Size: 8B
array(-854871110)
Coordinates:...
 DESIRED: 22865733
FAILED test/integration/test_integration_merge.py::test_merge_arrays[False] - AssertionError: 
Arrays are not almost equal to 7 decimals
 ACTUAL: <xarray.DataArray '__xarray_dataarray_variable__' ()> Size: 8B
array(-854871110)
Coordinates:
    spatial_ref  int64 8B 0
 DESIRED: 22865733
FAILED test/integration/test_integration_merge.py::test_merge__different_crs[True] - AssertionError: 
Arrays are not almost equal to 7 decimals
 ACTUAL: <xarray.DataArray '__xarray_dataarray_variable__' ()> Size: 8B
array(-231858673)
Coordinates:
    spatial_ref  int64 8B 0
 DESIRED: -131734881
FAILED test/integration/test_integration_merge.py::test_merge__different_crs[False] - AssertionError: 
Arrays are not almost equal to 7 decimals
 ACTUAL: <xarray.DataArray '__xarray_dataarray_variable__' ()> Size: 8B
array(-231858673)
Coordinates:
    spatial_ref  int64 8B 0
 DESIRED: -131734881
FAILED test/integration/test_integration_merge.py::test_merge_arrays__res - AssertionError: 
Arrays are not almost equal to 7 decimals
 ACTUAL: 3132355.0
 DESIRED: 13760565
```

PR #781, opened 2024-06-17
REF:merge: Use merge path & rio.to_raster

snowman2:
Related #765

There is quite a bit of logic in `rio.to_raster` that would make this more robust.


justingruca (approved):

snowman2:
Thanks @justingruca :+1:

PR #782, opened 2024-06-18
BUG: Raise overflow error when nodata conversion is unsafe

snowman2:
Changing to be consistent with Numpy 2: https://numpy.org/neps/nep-0050-scalar-promotion.html

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -3,6 +3,7 @@ History
 
 Latest
 ------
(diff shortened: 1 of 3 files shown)


snowman2:
Ignoring test failures due to new MacOS arm runners.

justingruca (approved):

snowman2:
Thanks @justingruca 👍

PR #784, opened 2024-06-19
DOC: multiple types

RichardScottOZ:
Not actually a possessive, something like this?

 - [ ] Closes #xxxx
 - [ ] Tests added


snowman2:
Thanks @RichardScottOZ 👍

PR #792, opened 2024-07-03
docs: fix minor code block issue for local installation in CONTRIBUTING.rst

dluks:
Fixes a typo which caused code block rendering to fail. Also adds quotes around `.[all]` as `pip install -e .[all]` fails in some environments. E.g.:
```console
$ python -m pip install -e .[all]
zsh: no matches found: .[all]

$ python -m pip install -e ".[all]"
... # (works)
```

:point_up: I didn't add anything to `docs/history.rst` as it doesn't appear to be the convention when updating documentation.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -78,9 +78,9 @@ Using python::
     $ python -m venv venv
     $ . venv/bin/activate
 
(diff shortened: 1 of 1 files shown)


snowman2:
Thanks @dluks :+1:

PR #793, opened 2024-07-03
MNT: recommend write_crs & deprecate set_crs

### LINKED ISSUE #743: Document difference between `set_crs` and `write_crs`
`set_crs`:

> Set the CRS value for the Dataset/DataArray without modifying the dataset/data array.

`write_crs`:

> Write the CRS to the dataset in a CF compliant manner.

The `input_crs` and `inplace` arguments are the same; the only difference in signature is that `write_crs` takes an optional `grid_mapping_name`.

In short, I'm confused :wink: 

Maybe consider `write_coordinate_system` in this issue as well.

dluks:
There is some confusion around when to use `write_crs` versus `set_crs`. This adds a small note to generally prefer write_crs when CRS persistence is needed. Fixes #743

 - [x] Closes #743 
 - [ ] Resolves all uncertainty around the initial issue(s)

One remaining nit that I would like to clarify in the documentation as well: when exactly _should_ people use `rio.set_crs()` and not `rio.write_crs()`?

### DIFF
--- a/docs/getting_started/crs_management.ipynb
+++ b/docs/getting_started/crs_management.ipynb
@@ -527,7 +527,9 @@
     "This modifies the `xarray.Dataset` or `xarray.DataArray` and sets the CRS in a CF compliant manner.\n",
     "\n",
     "- [rio.write_crs()](../rioxarray.rst#rioxarray.rioxarray.XRasterBase.write_crs)\n",
-    "- [rio.crs](../rioxarray.rst#rioxarray.rioxarray.XRasterBase.crs)"
+    "- [rio.crs](../rioxarray.rst#rioxarray.rioxarray.XRasterBase.crs)\n",
+    "\n",
+    "**Note:** It is generally recommended to use `rio.write_crs()` over `rio.set_crs()` if you want the CRS to persist on the Dataset/DataArray. Writing a Dataset/DataArray after calling only `rio.set_crs()` will not write the CRS information in a CF compliant manner."
    ]
   },
   {
@@ -1098,7 +1100,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.10.4"
+   "version": "3.12.4"
   }
  },
  "nbformat": 4,
--- a/docs/getting_started/crs_management.ipynb
+++ b/docs/getting_started/crs_management.ipynb
@@ -529,7 +529,7 @@
     "- [rio.write_crs()](../rioxarray.rst#rioxarray.rioxarray.XRasterBase.write_crs)\n",
     "- [rio.crs](../rioxarray.rst#rioxarray.rioxarray.XRasterBase.crs)\n",
     "\n",
-    "**Note:** It is generally recommended to use `rio.write_crs()` over `rio.set_crs()` if you want the CRS to persist on the Dataset/DataArray. Writing a Dataset/DataArray after calling only `rio.set_crs()` will not write the CRS information in a CF compliant manner."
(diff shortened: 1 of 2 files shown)


snowman2:
> when exactly should people use rio.set_crs() and not rio.write_crs()?

I would say never use `rio.set_crs` due to its instability. It was added a while a go and probably should be deprecated.

dluks:
@snowman2 I've pushed two commits that address the `set_crs` definition. The first adds a warning to the docstring and the second emits an actual `DeprecationWarning` when using `set_crs`. Not sure if raising an actual warning is what you had in mind, so if not I can revert it.

rioxarray/rioxarray.py
```diff
@@ -360,6 +361,9 @@ def set_crs(
         Set the CRS value for the Dataset/DataArray without modifying
         the dataset/data array.
 
+        DeprecationWarning: It is recommended to use `rio.write_crs()` instead. This
```
snowman2, line 364:
```suggestion
        .. deprecated:: 0.15.7
            It is recommended to use `rio.write_crs()` instead. This
```

rioxarray/rioxarray.py
```diff
@@ -372,6 +376,13 @@ def set_crs(
         :obj:`xarray.Dataset` | :obj:`xarray.DataArray`:
             Dataset with crs attribute.
         """
+        warnings.warn(
```
snowman2, line 379: I think raising a warning would be fine. However, `rio.set_crs` is used internally elsewhere. To avoid raising warnings where it is not needed, I suggest renaming this to `rio._set_crs` and use it internally. Then, create a new `rio.set_crs` that calls `rio._set_crs` after raising a warning.

docs/getting_started/crs_management.ipynb
```diff
@@ -527,7 +527,9 @@
     "This modifies the `xarray.Dataset` or `xarray.DataArray` and sets the CRS in a CF compliant manner.\n",
     "\n",
     "- [rio.write_crs()](../rioxarray.rst#rioxarray.rioxarray.XRasterBase.write_crs)\n",
-    "- [rio.crs](../rioxarray.rst#rioxarray.rioxarray.XRasterBase.crs)"
+    "- [rio.crs](../rioxarray.rst#rioxarray.rioxarray.XRasterBase.crs)\n",
+    "\n",
+    "**Note:** It is generally recommended to use `rio.write_crs()` over `rio.set_crs()` if you want the CRS to persist on the Dataset/DataArray. Writing a Dataset/DataArray after calling only `rio.set_crs()` will not write the CRS information in a CF compliant manner."
```
snowman2, line 532:
```suggestion
    "**Note:** It is recommended to use `rio.write_crs()` if you want the CRS to persist on the Dataset/DataArray and writes the CRS  CF compliant metadata. Calling only `rio.set_crs()` CRS storage method is lossy and will not modify the Dataset/DataArray metadata."
```

rioxarray/rioxarray.py
```diff
@@ -372,6 +376,13 @@ def set_crs(
...
+            "It is recommended to use `rio.write_crs()` instead. `rio.set_crs() will likely"
+            "be removed in a future release.",
+            DeprecationWarning,
```
snowman2, line 382:
```suggestion
            FutureWarning,
```

rioxarray/rioxarray.py
dluks: Good to know. I'll look into it.

snowman2:
Thanks @dluks :+1:

PR #796, opened 2024-07-05
ENH: Add allow_one_dimensional_raster option to rio.clip_box

### LINKED ISSUE #708: "At least one of the clipped raster x,y coordinates has only one point" Error when using rio.clip_box
Hi,

I am a beginner using the package to read rasters in a lazy mode. 

 I ran into this error message that seems should only occur when doing the integration test instead of general usage: "At least one of the clipped raster x,y coordinates has only one point." 

Is there any guideline that I can refer to solve this problem? Many thanks!

The function is here:
```
def raster_to_df_rioxarray(file_name, lb, rb, tb, bb):
    # read the raster file and convert to an xarray dataframe
    
    # try to open raster file with rioxarray, if nodata exception then return empty dataframe
    try: #print exception information if error happens and return empty dataframe
        raster  = rxr.open_rasterio(file_name).rio.clip_box( 
        minx = lb,
        miny = tb,
        maxx = rb,
        maxy = bb,
        ) 
    except Exception as e:
        # only print if e does not contain "No data", because these are the unexpected exceptions
        if 'No data' not in str(e):
            print('exception file name:', file_name, e)
        return None

    raster = raster.rename('intensity')

    df = raster.to_dask_dataframe()
    raster.close()
    df = df.where(df['intensity'] > 0)
    df = df.dropna()
    df = df.compute() # tranforming to pandas dataframe
    
    return df
```

snowman2:
- [x] Closes #708 
 - [x] Tests added

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -1,8 +1,9 @@
 History
 =======
 
-Latest
+0.16.0
 ------
+- ENH: Add `allow_one_dimensional_raster` option to `rio.clip_box` (issue #708)
 
 0.15.7
 ------
--- a/rioxarray/raster_array.py
+++ b/rioxarray/raster_array.py
@@ -713,10 +713,12 @@ class RasterArray(XRasterBase):
         auto_expand: Union[bool, int] = False,
         auto_expand_limit: int = 3,
         crs: Optional[Any] = None,
+        allow_one_dimensional_raster: bool = False,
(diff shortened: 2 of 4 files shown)


justingruca (approved):

snowman2:
Thanks @justingruca 👍

PR #797, opened 2024-07-18
REF:reproject: Make NaN default float nodata & update integer defaults

snowman2:
```python
OverflowError: Unable to convert nodata value (3.402823466e+38) to new dtype (float32).
```
Related #782

Integer defaults changed to match: https://github.com/rasterio/rasterio/blob/9e643c3f563a679aa5400d9b1a263df97b34f9e0/rasterio/dtypes.py#L99-L112

### DIFF
--- a/docs/history.rst
+++ b/docs/history.rst
@@ -3,6 +3,7 @@ History
 
 Latest
 ------
+- REF:reproject: Make NaN default float nodata & update integer defaults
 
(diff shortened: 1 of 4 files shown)


alfredoahds:
Looks good to me.

alfredoahds (commented):
Test review

PR #820, opened 2024-10-31
BUG: Fix chunk arguments for normalize_chunks

phofl:
- [ ] Closes #xxxx
 - [ ] Tests added

The current specification of the previous chunks is incorrect, it always needs to sum up to the whole axis if given as tuples. What you want is an integer here.

I've recently refactored how normalize_chunks works and this will break things for you. I'll add some compat code for now, but will want to remove this at some point, so fixing it over here

### DIFF
--- a/rioxarray/_io.py
+++ b/rioxarray/_io.py
@@ -929,7 +929,7 @@ def _prepare_dask(
             chunks=(1, "auto", "auto"),
             shape=(riods.count, riods.height, riods.width),
             dtype=_rasterio_to_numpy_dtype(riods.dtypes),
-            previous_chunks=tuple(c for c in block_shape),
+            previous_chunks=block_shape,
         )
     token_filename = filename
     if bidx is not None:
--- a/rioxarray/_io.py
+++ b/rioxarray/_io.py
@@ -929,7 +929,7 @@ def _prepare_dask(
             chunks=(1, "auto", "auto"),
             shape=(riods.count, riods.height, riods.width),
             dtype=_rasterio_to_numpy_dtype(riods.dtypes),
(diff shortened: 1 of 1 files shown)


phofl:
I am not sure how and if we can test this unfortunately.

rioxarray/_io.py
```diff
@@ -929,7 +929,7 @@ def _prepare_dask(
             chunks=(1, "auto", "auto"),
             shape=(riods.count, riods.height, riods.width),
             dtype=_rasterio_to_numpy_dtype(riods.dtypes),
-            previous_chunks=tuple((c,) for c in block_shape),
+            previous_chunks=tuple(c for c in block_shape),
```
snowman2, line 932: Would this work?
```suggestion
            previous_chunks=block_shape,
```

snowman2:
This was added in #31. I am concerned about modifying code that has been around for so long without some way to test and verify that the behavior works as expected.

phofl:
The previous way of specifying chunks was technically invalid, i.e. if you pass something like

```
chunks=((1, ), (1, ), (100, )) 
shape = (1, 10, 100)
```

to any array constructor (da.random.random for example), it will raise. The way this was intended was to pass (1, 1, 100) instead (historically, normalize_chunks treated both inputs equivalent). This PR addresses that.

Re testing: Every test in the test suite that uses ``chunks=True|"auto"`` tests this behaviour implicitly, so stuff would break if that would create an incompatible output.

I've put up a PR for Dask that makes both inputs equivalent for the time being and it's tested over there that both notations are equivalent.

snowman2:
After looking at dask's tests, I agree with your conclusion. Thanks!

phofl:
Great, thanks for reviewing!

PR #834, opened 2024-11-20
DOC: Update warning message for differing scales in rasterio attributes

### LINKED ISSUE #833: Different scales/offsets warning
While opening a geotiff with different scales I got this warning:
https://github.com/corteva/rioxarray/blob/master/rioxarray/_io.py#L673
```
    if len(set(riods.scales)) > 1:
        attrs["scales"] = riods.scales
        warnings.warn(
            "Offsets differ across bands. The 'scale_factor' attribute will "
            "not be added. See the 'scales' attribute."
        )
```
It was a bit confusing as the offsets were all 0s. Might be clearer if the warning would be "Scales differ across bands..."

dzanaga:
- [x] Closes #833

### DIFF
--- a/rioxarray/_io.py
+++ b/rioxarray/_io.py
@@ -670,7 +670,7 @@ def _get_rasterio_attrs(riods: RasterioReader):
     if len(set(riods.scales)) > 1:
         attrs["scales"] = riods.scales
         warnings.warn(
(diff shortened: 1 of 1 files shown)


snowman2:
Thanks @dzanaga 👍

Issue #588, opened 2022-10-14
Experimentally support CRSIndex


dcherian:
It'd be nice to experiment some with this prototype [CRSIndex](https://github.com/dcherian/crsindex/blob/main/crsindex.ipynb) (though there are still [bugs](https://github.com/pydata/xarray/issues/7162) and [work to do](https://github.com/pydata/xarray/issues/6293))

We could experiment with optionally adding such an index through `rioxarray.open_dataset(..., use_crsindex=True)` and then get some experience with it before refactoring it out to a ecosystem package like `geoxarray`

Reprojection seems like a high value application where assigning a new CRSIndex would be quite impactful.

Is there an existing [example](https://corteva.github.io/rioxarray/stable/examples/examples.html) that might benefit a lot from propagating CRS information automatically?

cc @djhoese

snowman2:
Thanks for making the prototype. That looks pretty nice.

Do you think it would be better to wait before officially adding it to rioxarray so some wrinkles can be ironed out first? Or do you think that adding it to rioxarray will help the development process for indexes in xarray?

Thoughts:
- Add the `CRSIndex` in `rioxarray` in the `rio.write_crs` method. That is used everywhere and should propagate to all of the `rioxarray` methods that use the `CRS`.
- As an initial implementation, we could start with warnings instead of errors. This way, we aren't preventing operations from working that should work if the user really knows what they are doing. Later on, as we gain more experience with indexes and feel more confident in the implementation, we can update to exceptions.
- We may want to ignore axis order when comparing the CRS with the [ignore_axis_order kwarg](https://pyproj4.github.io/pyproj/stable/api/crs/crs.html#pyproj.crs.CRS.equals).

dcherian:
> adding it to rioxarray will help the development process for indexes in xarray?

I think this will be a very useful experiment.

> dd the CRSIndex in rioxarray in the rio.write_crs method. That is used everywhere and should propagate to all of the rioxarray methods that use the CRS.

IMO it would be good to try this and see what fails in the test suite. That would be a more exhaustive test than my suggestion to pick a single "workflow" example./

snowman2:
Sounds like a good plan to me.

Anyone who would like to try implementing the CRSIndex and adding it to `rio.write_crs` is welcome to, If you do get started, please post a comment here so we can coordinate our efforts. I will make sure to post in the thread if I get some free time to work on it.

scottyhq:
[...]

PR #846, opened 2025-02-20
Add raster index

benbovy:
- [ ] Tests added

This work in progress PR adds a custom `RasterIndex` Xarray index that is built on top of https://github.com/pydata/xarray/pull/9543 (available in Xarray's main branch).

- Opt-in feature via `rioxarray.set_option(use_raster_index=True)`
  - Alternatively or in addition we could expose this option as an IO backend option. The global option is handy for the other places where new spatial coordinates are generated (e.g., reproject).
- `RasterIndex` should work with both x/y independent axis 1D coordinates (rectilinear affine transform with no rotation) vs. x/y dependent axis 2D coordinates. In the latter case data selection is more limited.
- The spatial coordinates freshly generated with a `RasterIndex` are lazy (in both 1D and 2D cases): they wrap the affine transform.
- There are still quite a few things to implement and test.
- I skipped implementing CRS-aware functionality here (#588) as I'm focusing on the coordinate transform (also this PR is already getting big), but I guess this could be added later on top of this.

TODO on the Xarray Index API implementation (could also be done after this PR, if needed):

- [ ] `RasterIndex.from_variables`: compute the affine transformation from explicit coordinate values, although maybe not very useful?
- [ ] `RasterIndex.concat`: possible to compute a new affine transformations by combining the transformations of multiple raster indexes?
- [ ] `RasterIndex.join`: like for `concat` we could perhaps combine the transformations in some simple cases (check that the two affine transforms are compatible)?
- [ ] `RasterIndex.reindex_like`: maybe this could be implemented using a forward transformation of the current index chained with a reverse transformation of the other index?
- [ ] `RasterIndex.rename`: must propagate the new coordinate and/or dimension names to the underlying `AffineTransform` objects!
- [ ] `RasterIndex.copy` (not sure we need to override Index.copy)
- [ ] `RasterIndex.__getitem__` (not sure we really need to implement it)

Here is an [example notebook](https://notebooksharing.space/view/2e7ed956213d137443ebacf5fabecaf5e772e642d6a7ac12513575d1ebd98fce#displayOptions=) that I'll update as I continue working on this PR.

There are a few issues that we need to fix in Xarray so this PR can work properly (I've added a few temporary workarounds when possible):

- https://github.com/pydata/xarray/pull/10066
- https://github.com/pydata/xarray/pull/10067

cc @scottyhq @dcherian @maxrjones

### DIFF
--- a/rioxarray/raster_index.py
+++ b/rioxarray/raster_index.py
@@ -2,12 +2,14 @@ from collections.abc import Hashable, Mapping
 from typing import Any
 
 import numpy as np
+import pandas as pd
 from affine import Affine
 from xarray import DataArray, Index, Variable
+from xarray.core.coordinate_transform import CoordinateTransform
+
 # TODO: import from public API once it is available
 from xarray.core.indexes import CoordinateTransformIndex, PandasIndex
 from xarray.core.indexing import IndexSelResult, merge_sel_results
-from xarray.core.coordinate_transform import CoordinateTransform
 
 
 class AffineTransform(CoordinateTransform):
@@ -27,7 +29,9 @@ class AffineTransform(CoordinateTransform):
         y_dim: str = "y",
         dtype: Any = np.dtype(np.float64),
     ):
-        super().__init__((x_coord_name, y_coord_name), {x_dim: width, y_dim: height}, dtype=dtype)
+        super().__init__(
+            (x_coord_name, y_coord_name), {x_dim: width, y_dim: height}, dtype=dtype
+        )
         self.affine = affine
 
         # array dimensions in reverse order (y = rows, x = cols)
@@ -78,7 +82,7 @@ class AxisAffineTransform(CoordinateTransform):
         is_xaxis: bool,
         dtype: Any = np.dtype(np.float64),
     ):
-        assert (affine.is_rectilinear and (affine.b == affine.d == 0))
+        assert affine.is_rectilinear and (affine.b == affine.d == 0)
 
         super().__init__((coord_name,), {dim: size}, dtype=dtype)
         self.affine = affine
@@ -113,9 +117,13 @@ class AxisAffineTransform(CoordinateTransform):
 
         # only compare the affine parameters of the relevant axis
         if self.is_xaxis:
-            affine_match = self.affine.a == other.affine.a and self.affine.c == other.affine.c
+            affine_match = (
+                self.affine.a == other.affine.a and self.affine.c == other.affine.c
+            )
         else:
-            affine_match = self.affine.e == other.affine.e and self.affine.f == other.affine.f
+            affine_match = (
+                self.affine.e == other.affine.e and self.affine.f == other.affine.f
+            )
 
         return affine_match and self.size == other.size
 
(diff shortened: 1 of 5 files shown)


rioxarray/raster_index.py
```diff
@@ -0,0 +1,428 @@
...
+        return merge_sel_results(results)
+
+    def equals(self, other: Index) -> bool:
+        if not isinstance(other, RasterIndex):
+            return False
+        if set(self._wrapped_indexes) != set(other._wrapped_indexes):
+            return False
+
+        return all(
+            index.equals(other._wrapped_indexes[k])  # type: ignore[arg-type]
```
benbovy, line 407: Need to relax the typing constraints of `xarray.core.indexes.CoordinateTransformIndex.equals()` (other should be any `xarray.Index`) so we can remove the `# type: ignore` here.

rioxarray/raster_index.py
```diff
@@ -0,0 +1,428 @@
...
+from affine import Affine
+from xarray import DataArray, Index, Variable
+from xarray.core.coordinate_transform import CoordinateTransform
+
+# TODO: import from public API once it is available
+from xarray.core.indexes import CoordinateTransformIndex, PandasIndex
+from xarray.core.indexing import IndexSelResult, merge_sel_results
+
+
+class AffineTransform(CoordinateTransform):
```
dcherian, line 16: I think the `Affine*` classes should live in Xarray. They are useful for both geo and bio-imaging domains at least (and possibly astro too).

benbovy: Yes I agree these are useful in many domains! The wrapped `affine.Affine` object is here restricted to 2D affine transformation of the plane so it might be limited for general usage, though. These classes are also thin wrappers.

rioxarray/raster_index.py
```diff
@@ -0,0 +1,428 @@
...
+    axis_transform: AxisAffineTransform
+    dim: str
+
+    def __init__(self, transform: AxisAffineTransform):
+        assert isinstance(transform, AxisAffineTransform)
+        super().__init__(transform)
+        self.axis_transform = transform
+        self.dim = transform.dim
+
+    def isel(  # type: ignore[override]
```
benbovy, line 196: Again here we should make the return type of `xarray.core.Index.isel()` more permissive (i.e., any `Index` object or `None`) so we can remove the `# type: ignore` comment.

rioxarray/raster_index.py
```diff
@@ -0,0 +1,428 @@
...
+                new_index = index.isel(index_indexers)
+                if new_index is not None:
+                    new_indexes[coord_names] = new_index
+
+        if new_indexes:
+            # TODO: if there's only a single PandasIndex can we just return it?
+            # (maybe better to keep it wrapped if we plan to later make RasterIndex CRS-aware)
+            return RasterIndex(new_indexes)
+        else:
+            return None
```
benbovy, line 382: When it is possible, it would be convenient to return a PandasIndex instead of a RasterIndex for reusing the selection results in further Xarray operations that require alignment with other Xarray objects. The reason is that Xarray alignment is based on strict Index types, e.g., it cannot compare or align a RasterIndex (encapsulating a single PandasIndex) with another PandasIndex.

However, we can only return a PandasIndex in a limited number of cases, e.g.,

- it is possible for `da.isel(x=[...], y=0)` (or `da.isel(x=0, y=[...])`) to return a "x" (or "y") coordinate with a PandasIndex and to drop the index for the "y" (or "x") scalar coordinate.
- it is not possible for `da.isel(x=[...], y=[...])` to return both "x" and "y" coordinates each with a PandasIndex. This is because `xarray.Index.isel()` doesn't currently allow returning multiple index objects. 

Returning a PandasIndex in some cases and a RasterIndex in other cases would make the behavior complex and pretty confusing.

Alternatively, we could always return a RasterIndex for both "x" and "y" coordinates even when selecting it with scalar values. This would provide simple behavior and might facilitate CRS-aware operations even with scalar spatial coordinates, but this would be even less Xarray-alignment friendly.

So in short: it is not straightforward to choose the right behavior here. It is a trade-off.

rioxarray/raster_index.py
dcherian: They are "thin" in code but 'deep' in understanding of the Xarray model. That's why I think we should vendor it. Even so, vendoring will save a bunch of your time as you iterate on this since you wouldn't need to keep code in two places in-sync.

benbovy: Maybe the whole content of `raster_index.py` including the `RasterIndex` class could (eventually) be moved into Xarray or some dedicated package? There's nothing IO specific and it could be useful in other downstream packages like odc-geo, geoxarray, etc.

rioxarray/raster_index.py
dcherian: >  but this would be even less Xarray-alignment friendly.

I think this is fine. Users can `reset_index` once they are sure what they're about to do makes sense

snowman2:
This has been an in-progress MR for a while. Any plans to keep working on it?

dcherian:
Yes, we landed on keeping the index separate since it's generally useful outside just rioxarray. Are you ok with adding another dependency for this feature?

snowman2:
> A few of us are meeting Monday morning 4/21 8.30 MT. Would you like to join?

Sure, I would be happy to join.

PR #855, opened 2025-04-16
Generate spatial coords with rasterix.RasterIndex

benbovy:
- [x] Supersedes #846 
 - [ ] Tests added

Optionally generates the spatial x/y coordinates with a [rasterix](https://github.com/dcherian/rasterix)'s `RasterIndex` (opt-in).

The [example notebook](https://notebooksharing.space/view/2e7ed956213d137443ebacf5fabecaf5e772e642d6a7ac12513575d1ebd98fce#displayOptions=) linked in #846 should work here as well.

### DIFF
--- a/rioxarray/rioxarray.py
+++ b/rioxarray/rioxarray.py
@@ -121,23 +121,39 @@ def affine_to_coords(
 
 def _generate_spatial_coords(
     *, affine: Affine, width: int, height: int
-) -> dict[Hashable, Any]:
+) -> xarray.Coordinates:
     """get spatial coords in new transform"""
-    new_spatial_coords = affine_to_coords(affine, width, height)
-    if new_spatial_coords["x"].ndim == 1:
-        return {
-            "x": xarray.IndexVariable("x", new_spatial_coords["x"]),
-            "y": xarray.IndexVariable("y", new_spatial_coords["y"]),
-        }
-    return {
-        "xc": (("y", "x"), new_spatial_coords["x"]),
-        "yc": (("y", "x"), new_spatial_coords["y"]),
-    }
+
+    if get_option("use_raster_index"):
+        try:
+            from rasterix import RasterIndex
+        except (ImportError, ModuleNotFoundError) as err:
+            raise ImportError(
+                "The rasterix package is needed for generating "
+                "the spatial coordinates with a RasterIndex"
+            ) from err
+
+        raster_index = RasterIndex.from_transform(affine, width, height)
+        return xarray.Coordinates.from_xindex(raster_index)
+
+    else:
+        new_spatial_coords = affine_to_coords(affine, width, height)
+        if new_spatial_coords["x"].ndim == 1:
+            coords = {
+                "x": xarray.IndexVariable("x", new_spatial_coords["x"]),
+                "y": xarray.IndexVariable("y", new_spatial_coords["y"]),
+            }
+        else:
+            coords = {
+                "xc": (("y", "x"), new_spatial_coords["x"]),
+                "yc": (("y", "x"), new_spatial_coords["y"]),
+            }
+        return xarray.Coordinates(coords)
 
 
 def _get_nonspatial_coords(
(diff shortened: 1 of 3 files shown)


benbovy:
Yes I will do all of that before marking this PR as ready.

> Does the xarray pin need to be updated?

Likely for the tests yes. Since this is an opt-in feature Rioxarray may (should) still support Xarray versions older than the minimum version required by Rasterix.

rioxarray/rioxarray.py
```diff
@@ -121,23 +121,39 @@ def affine_to_coords(
...
+    if get_option("use_raster_index"):
+        try:
+            from rasterix import RasterIndex
+        except (ImportError, ModuleNotFoundError) as err:
+            raise ImportError(
+                "The rasterix package is needed for generating "
+                "the spatial coordinates with a RasterIndex"
+            ) from err
+
+        raster_index = RasterIndex.from_transform(affine, width, height)
```
dcherian, line 136: I think we should also delete the transform from the grid_mapping variable like so:
```python
grid_mapping = ds.rio.grid_mapping
obj[grid_mapping].attrs.pop("GeoTransform")
```

following https://github.com/dcherian/rasterix/blob/main/design_notes/raster_index.md#read-time and our discussions. This will trigger some more changes (`_cached_transform` will need to pull the Affine from RasterIndex)

dcherian:
How should we bring in `xproj` to handle CRS here? A followup PR?

benbovy:
> How should we bring in xproj to handle CRS here? A followup PR?

Yes, a separate PR will be easier I think. Interoperability between xproj and rioxarray would likely be handled [at the accessor level](https://xproj.readthedocs.io/en/latest/integration.html#crs-aware-xarray-accessor) and might require some more refactoring.

rioxarray/rioxarray.py
```diff
@@ -162,7 +178,7 @@ def _get_nonspatial_coords(
                 src_data_array[coord].values,
                 src_data_array[coord].attrs,
             )
-    return coords
+    return xarray.Coordinates(coords)
 
 
 def _make_coords(
```
dcherian, line 184: It'd be nice to have an accessor function to set this index even if I opened a Zarr file for example. Is there something like `ds.rio.assign_coords`? I didn't see anything similar in the docs. Perhaps we should have `ds.rio.assign_indexes()` that does both RasterIndex and CRSIndex

snowman2: Do you think that would be a good fit to add to a rasterix accessor? If you do that, you just set `decode_coords=False` and assign the index/coords post load.

dcherian: It needs a bunch of extra info : x dim name, y dim name, transform, and so it's pretty convenient to just do all that with the `.rio` accessor. It seems like a prudent and convenient addition

snowman2: Inside rasterix, it could look for the `.rio` , `.odc`, or `.geo` accessors inside the method. It could be done similar to `open_dataset` where it automatically figures out how to open files based on what's available. It also allows the user select their preferred engine. If they don't exist, then raise an error.

dcherian: That is an option but to me it feels a lot cleaner to add a method under the `rio` namespace. (and similarly under the other namespaces if they thought it was a good idea). At least for rioxarray it's just a small wrapper around an existing function, so it doesn't seem like meaningful extra maintenance burden to me.

benbovy: > Inside rasterix, it could look for the .rio , .odc, or .geo accessors inside the method

Hmm this wouldn't be ideal IMHO, assuming that the goal of the rasterix (and xproj) project is to provide reusable utilities to downstream packages like rioxarray, odc-geo, etc. even though technically looking for those accessors is possible without having cyclic dependencies.

Under the same assumption, rasterix and xproj should also be flexible / not strongly opinionated about the names of the spatial coordinates and dimensions, etc.

snowman2:
Closing due to inactivity. Implementation planning in #908.

PR #879, opened 2025-12-08
Convert tuple chunks to dict in `_prepare_dask` to avoid DeprecationWarning

### LINKED ISSUE #878: Providing chunks as tuples deprecated by xarray
Related to #253.

As of [xarray 2023.11.0](https://github.com/pydata/xarray/pull/8341), providing chunksizes as dimension-order tuples is deprecated.

In rioxarray, in places chunks are still provided as tuples to xarray functions. One of the more common use cases is when passing "auto" or True as the `chunks` argument to `open_rasterio`, as `_prepare_dask` will turn those into tuples.

To reproduce:
```python
import warnings
warnings.simplefilter("error", DeprecationWarning)

import rioxarray
ds = rioxarray.open_rasterio("file.tif", chunks=True)
```

If a user actually provides a tuple for the `chunks` argument, then the warning makes sense. But ideally passing "auto" or True shouldn't cause such warnings. Perhaps `_prepare_dask` could be adapted to use a dict instead of tuple for the "auto"/True case?

martijnvandermarel:
- [x] Closes #878 
 - [x] Tests added

Providing chunks as tuples to `open_rasterio` is still allowed (and doesn't generate a warning) so the interface remains the same. However in `_prepare_dask` any chunks provided as tuples are converted to a dict, assuming the dimension ordering in the provided tuple is (band, y, x) (the same as the returned dataarray). If the tuple is not of length 3 an error is raised.

This should remove the DeprecationWarning by xarray when passing chunks=True, chunks="auto" or chunks as a tuple.

### DIFF
--- a/rioxarray/_io.py
+++ b/rioxarray/_io.py
@@ -931,8 +931,6 @@ def _prepare_dask(
             dtype=_rasterio_to_numpy_dtype(riods.dtypes),
             previous_chunks=block_shape,
         )
-
-    if isinstance(chunks, tuple):
         # xarray wants chunks as a dict rather than a tuple
         chunks = dict(zip(result.dims, chunks, strict=True))
 
--- a/test/integration/test_integration__io.py
+++ b/test/integration/test_integration__io.py
@@ -754,8 +754,8 @@ def test_chunks():
 
 
 @pytest.mark.filterwarnings("error::DeprecationWarning")
-@pytest.mark.parametrize("chunks", [True, (1, 100, 100)])
-def test_chunks_as_tuple_no_deprecation_warning(chunks):
+@pytest.mark.parametrize("chunks", [True, "auto"])
+def test_auto_chunks_no_deprecation_warning(chunks):
     with rioxarray.open_rasterio(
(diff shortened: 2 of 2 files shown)


rioxarray/_io.py
snowman2, line 927: I wonder if instead you could pass in a dictionary?

```python
chunks = {
    result.rio.x_dim: "auto",
    result.rio.y_dim: "auto",
}
if extra_dim := result.rio._check_dimensions():
    chunks[extra_dim] = 1
```

martijnvandermarel: `normalize_chunks` always returns a tuple, even if the input chunks are a dict. Or are you suggesting to not use `normalize_chunks` at all, and directly pass the `chunks` dict `{result.rio.x_dim: "auto", result.rio.y_dim: "auto", extra_dim: 1}` into the `result.chunk` method?

snowman2: Looks like you are correct. I didn't dig into it and was incorrectly assuming the input and output format would be the same.

snowman2:
Thanks @martijnvandermarel 👍

PR #883, opened 2025-12-12
Implement Zarr spatial and proj conventions support

emmanuelmathot:
Draft for Zarr conventions support according to plan in #882

 - [x] Closes #882
 - [x] Tests added
 - [x] Fully documented, including `docs/history.rst` for all changes and `docs/rioxarray.rst` for new API

PR #899, opened 2026-01-28
Add Convention option & extract CF logic to cf.py

emmanuelmathot:
This is the first in a series of PRs splitting #883 as requested by maintainers.

## Summary
This PR establishes the framework for supporting multiple geospatial metadata conventions. Reading auto-detects conventions from existing attributes, while writing uses the convention setting.

### New Convention Framework
- Add `Convention` enum with `CF` value in `rioxarray/enum.py`
- Add `convention` option to `set_options()` for controlling write behavior
- Create `ConventionProtocol` interface for consistent convention module APIs

### CF Convention Module (`rioxarray/_convention/cf.py`)
- Extract CF-specific read/write logic from `rioxarray.py`
- `read_crs()` - Read CRS from CF grid_mapping coordinate
- `read_transform()` - Read transform from GeoTransform attribute
- `read_spatial_dimensions()` - Detect spatial dimensions from standard names/attrs
- `write_crs()` - Write CRS with CF conventions (handles `grid_mapping_name` defaults)
- `write_transform()` - Write transform with CF conventions
- `_write_grid_mapping()` - Internal helper for grid_mapping encoding

### Auto-detect Layer (`rioxarray/_convention/__init__.py`)
- `read_crs_auto()` - Try all convention readers, with legacy `attrs['crs']` fallback
- `read_transform_auto()` - Try all convention readers, with legacy `attrs['transform']` fallback  
- `read_spatial_dimensions_auto()` - Try all convention readers
- `_get_convention()` - Get convention module based on setting

### Refactored `rioxarray.py`
- `crs` property now uses `read_crs_auto()` for auto-detection
- `_cached_transform()` now uses `read_transform_auto()`
- `write_crs()` delegates to convention module, accepts `convention` parameter
- `write_transform()` delegates to convention module, accepts `convention` parameter
- `write_grid_mapping()` method kept for backwards compatibility
- Spatial dimension detection uses `read_spatial_dimensions_auto()`

### API Design Principles
- **Reading**: Auto-detects conventions by trying each reader until one succeeds
- **Writing**: Uses global `convention` setting (default: CF) or per-method `convention` parameter
- **Backwards compatible**: Legacy `crs`/`transform` attrs still supported for reading
- **Keyword-only arguments**: All CF functions use `*,` for clarity and API stability

## Files Changed
- `rioxarray/enum.py` - New `Convention` enum
- `rioxarray/_options.py` - Add `convention` option
[...]

PR #900, opened 2026-01-28
Add read support for Zarr spatial and proj conventions

emmanuelmathot:
This is the second in a series of PRs splitting #883 as requested by maintainers. Depends on #899.

**Changes:**
- Add `Convention.Zarr` enum value
- Create `_convention/zarr.py` module with read functions for Zarr conventions
- Update rioxarray.py to try both conventions when reading (CF first by default, Zarr first with `set_options(convention=Convention.Zarr)`)
- Support `proj:wkt2`, `proj:code`, `proj:projjson` for CRS reading
- Support `spatial:transform`, `spatial:dimensions` for transform and dimension reading

**Reviewer feedback addressed:**
- ✅ Both conventions are attempted when reading regardless of setting (priority changes only)
- ✅ Use `CRS.from_user_input()` instead of explicit PROJJSON handling
- ✅ Use functional test format, not class format
- ✅ Imports at top of file

---

- [ ] Closes #883 (partial - second of series)
- [x] Tests added
- [X] Fully documented, including history.rst for all changes and rioxarray.rst for new API

### DIFF
--- a/rioxarray/_convention/zarr.py
+++ b/rioxarray/_convention/zarr.py
@@ -1,0 +1,326 @@
+"""
+Zarr spatial and proj convention support for rioxarray.
+
+This module provides functions for reading geospatial metadata according to:
+- Zarr spatial convention: https://github.com/zarr-conventions/spatial
+- Zarr geo-proj convention: https://github.com/zarr-experimental/geo-proj
+"""
+import json
+from typing import Optional, Union
+
+import rasterio.crs
+import xarray
+from affine import Affine
+
+from rioxarray.crs import crs_from_user_input
+
+# Convention identifiers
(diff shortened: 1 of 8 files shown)


rioxarray/_convention/zarr.py
```diff
@@ -0,0 +1,326 @@
...
+
+    Returns
+    -------
+    rasterio.crs.CRS or None
+        Parsed CRS object, or None if not found
+    """
+    if convention_check and not has_convention_declared(attrs, "proj:"):
+        return None
+
+    for proj_attr, parser in [
```
snowman2, line 207: `crs_from_user_input` should work for all if the inputs. No need for individual parsers.

PR #901, opened 2026-01-28
Propagate rasterio's meta and profile

remi-braun:
- [x] Closes #804
 - [x] Tests added


remi-braun:
No real need when you are familiar with rioxarray, you're right.
Maybe just adding some documentation about the equivalence between the two libs should be enough.
I'll try to draft something in the next days.
You can close this if your comments also apply to `profile` 👍

PR #925, opened 2026-05-04
DOC: Fix various documentation errors and warnings

djhoese:
I've been running into a weird issue when trying to document my own project that imports rioxarray. While looking into it I found that there are a ton of warnings and errors in rioxarray's own documentation rendering so I fixed all the ones I was hitting locally:

1. sphinx "language" in conf.py needs a real value now. I went with "en".
2. Inconsistent title levels in `crs_management.ipynb`. Basically the titles went from level 2 to level 4 with no level 3 in between. Further "API Documentation" was defined as level 4 but really I think it should be on the same level as the other sections at 3. I fixed what I could find. This is actually a really big deal as these titles just didn't exist in the rendered HTML because of these issues.
3.  Various indentation or bullet type inconsistencies.
4. Footnotes that were never referenced so switched to bullet points.
5. Cases of `Union[Resampling | str]`. This is redundant as the `|` is a Union. I changed this to match the other cases in the modules with `Union[Resampling, str]` (note the comma instead of the vertical bar). This is the issue I was running into so I'm hoping once this is merged and released that this will just work for my own project.

 - [ ] Closes #xxxx
 - [ ] Tests added


djhoese:
Test failures seem unrelated to my changes.

snowman2:
Thanks @djhoese 👍

djhoese:
I realized I could just hack the rioxarray source in my installed environment and I can confirm that the changes to `Union[Resampling | str]` have removed these warnings:

```
WARNING: Failed to import <my_module>.
Possible hints:
* AttributeError: module <my_module> has no attribute 'some_attr'
* TypeError: unsupported operand type(s) for |: 'Resampling' and 'type'
```

PR #927, opened 2026-06-08
MNT: several updates to pyproject.toml, ignore uv.lock

mwtoews:
Apologies for cobbling a few fixes into one PR.

These changes are related to [PEP 639](https://peps.python.org/pep-0639/), and clear up deprecation warnings when building:
- Change `[project.license]` from a table to a SPDX license expression
- Add `[project.license-files]`
- Remove the license from the classifiers
- Correction `[options.package_data]` -> `[tool.setuptools.package-data]`

Other changes:
- Remove `[tool.black]`, because `target_version` was incorrect (should be `target-version`), but more importantly this was tracking an old version of Python, and by default, Black will infer target versions from the project metadata ([docs](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#t-target-version))
- Ignore `uv.lock` files, which are created with ([e.g.](https://docs.astral.sh/uv/concepts/projects/sync/)) `uv sync --all-extras`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,6 +83,9 @@ celerybeat-schedule
 # dotenv
 .env
 
+# uv env
+uv.lock
+
 # virtualenv
 .venv
 venv/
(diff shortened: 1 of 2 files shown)


snowman2:
Thanks @mwtoews 👍

